PR #105, opened 2015-01-13
Made references to nuget packages in project files relative to solution.

smischke:
This makes it easier to include the projects in other solutions. It saves me some work.


objorke:
Will these changes be reverted if we do a NuGet package update in the solutions in this project?

smischke:
They won't be reverted by automatic or manual package restore, but they will be reverted when updating to a new version of the package. Tested with nunit:
<HintPath>$(SolutionDir)packages\NUnit.2.6.3\lib\nunit.framework.dll</HintPath>
becomes
<HintPath>..\packages\NUnit.2.6.4\lib\nunit.framework.dll</HintPath>

I can understand if you don't want that. The relevant files for me are just:
HelixToolkit_NET40.csproj
HelixToolkit.Wpf_NET40.csproj
HelixToolkit.Wpf.SharpDX_NET40.csproj

Not too much work for me. Forget about this pull request. :)

Issue #106, opened 2015-01-14
System.InvalidOperationException on large scale zooming


chrkon:
In our Application we get an InvalidOperationException if we zoom out to a very large scale. This happens easiely if you you use the spacenavigator mouse. 
We receive the following error message:
```
System.InvalidOperationException wurde nicht von Benutzercode behandelt.
HResult=-2146233079
Message=Die Matrix kann nicht invertiert werden, da die Matrix nicht invertierbar ist.
Source=PresentationCore
StackTrace:
bei System.Windows.Media.Media3D.Matrix3D.Invert()
bei HelixToolkit.Wpf.Matrix3DExtensions.Inverse(Matrix3D m) in c:\projects\helix-toolkit\Source\HelixToolkit.Wpf\ExtensionMethods\Matrix3DExtensions.cs:Zeile 33.
bei HelixToolkit.Wpf.ScreenGeometryBuilder.UpdateTransforms() in c:\projects\helix-toolkit\Source\HelixToolkit.Wpf\Helpers\ScreenSpace\ScreenGeometryBuilder.cs:Zeile 95.
bei HelixToolkit.Wpf.PointsVisual3D.UpdateTransforms() in c:\projects\helix-toolkit\Source\HelixToolkit.Wpf\Visual3Ds\ScreenSpaceVisuals\PointsVisual3D.cs:Zeile 88.
bei HelixToolkit.Wpf.ScreenSpaceVisual3D.OnCompositionTargetRendering(Object sender, RenderingEventArgs e) in c:\projects\helix-toolkit\Source\HelixToolkit.Wpf\Visual3Ds\ScreenSpaceVisuals\ScreenSpaceVisual3D.cs:Zeile 217.
bei HelixToolkit.Wpf.WeakEventListener`2.System.Windows.IWeakEventListener.ReceiveWeakEvent(Type managerType, Object sender, EventArgs e) in c:\projects\helix-toolkit\Source\HelixToolkit.Wpf\Helpers\WeakEventListeners\WeakEventListener
{TEventManager,TEventArgs}

.cs:Zeile 49.
bei System.Windows.WeakEventManager.ListenerList.DeliverEvent(Listener& listener, Object sender, EventArgs args, Type managerType)
bei System.Windows.WeakEventManager.ListenerList.DeliverEvent(Object sender, EventArgs args, Type managerType)
bei System.Windows.WeakEventManager.DeliverEventToList(Object sender, EventArgs args, ListenerList list)
[...]

PR #107, opened 2015-01-14
#106 Bugfix in ScreenGeometryBuilder. Added a check for "hasInverse".

chrkon:
Regarding to Issue #106 I found a missing check for "hasInverse" before the ProjectionToScreen matrix is inverted. After adding this check to the helix toolkit our application did not crash with this error anymore.

Kind regards,
Christof Konstantinopoulos


objorke:
Looks good! Thanks!

Issue #85, opened 2014-11-18
Issue when restoring SharpDX-version from taskbar


Isdas:
Hi,
If I minimize one of the SharpDX-demos and start working on another program (i.e. Word / Excel) I sometimes have problem restoring the demo. I get a "pling"-sound when I click on the icon, but nothing happens. This can also be triggered when clicking fast on the icon.

I found a simular issue when a person used XNA:
http://stackoverflow.com/questions/8539158/window-emitting-a-sound-when-i-try-to-restore-it-from-the-taskbar

Anyone got an idea how to fix this? I've tried several things :). "Something" is eating up the event!

This only happens in the SharpDX-version.

Best regards
-Isdas

objorke:
Is this solved by #112?

objorke:
I understand this problem is still not resolved. Help wanted!

smischke:
Oystein, can you reproduce the problem? If so, what are the steps?

objorke:
Sorry, I have not tried. Erik, can you provide details?

Isdas:
Well, to be able to not restore the application from the taskbar you either have to have a slow PC or a "large" application on top of the engine (SharpDX-version). Personally I use a standard DevExpress ribbon application where I have the engine inside one of the pages.

The way I reproduce the problem is to minimize the program and then start to use other application (Word, Excel, VS, etc.) and then try to maximize the program again. You then get a "pling" sound indicating that something is wrong.

The only way I have managed to reproduce the problem with the example-codes in the SharpDX version is to click very fast on the icon (minimizing and maximizing the program). If you do this you also get this "pling"-sound. This does not happen in the non-SharpDX-version (same procedure).

The link above states the same problem that I am experiencing. It is a very "odd" bug I must say, but for the ones using our program it can be a bit problematic (usually they force shutdown on the application since they can't restore it).

I have googled this problem alot and tried many different workarounds; without luck. I THINK it is a threading problem, and I would guess that one thread is ruining everything (or maybe a dispatcher). This can maybe be resolved by thread priorities, but I haven't manged to succeed in this way either.

Please let me know if you find out something or need more information :)

-Isdas

holance:
Should fixed now.

PR #112, opened 2015-01-16
Fixed an issue on NET40 where the render targets get lost and not restor...

smischke:
...ed after UAC popup or sleep / hibernate.


smischke:
Might also solve issue 85.
https://github.com/helix-toolkit/helix-toolkit/issues/85

PR #120, opened 2015-02-12
Prevent crash if no DX10 GPU is present (VMWare)

smischke:
This small fix will prevent a crash if no DX10 GPU is present (like when running in VMWare). I also have a plan for MVVM friendly error handling, but this is a first step.

### DIFF
--- a/Source/HelixToolkit.Wpf.SharpDX/Controls/DPFCanvas.cs
+++ b/Source/HelixToolkit.Wpf.SharpDX/Controls/DPFCanvas.cs
@@ -146,8 +146,8 @@ namespace HelixToolkit.Wpf.SharpDX
         public DPFCanvas()
         {
             this.renderTimer = new Stopwatch();
-            this.Loaded += this.WindowLoaded;
-            this.Unloaded += this.WindowClosing;
+            this.Loaded += this.OnLoaded;
+            this.Unloaded += this.OnUnloaded;
             this.ClearColor = global::SharpDX.Color.Gray;
             this.IsShadowMapEnabled = false;            
             this.IsMSAAEnabled = true;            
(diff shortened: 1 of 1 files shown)


Source/HelixToolkit.Wpf.SharpDX/Controls/DPFCanvas.cs
```diff
@@ -180,15 +180,29 @@ public void InvalidateRender()
...
+            {
+                this.StartD3D();
+                this.StartRendering();
+            }
+            catch (Exception)
+            {
+                // Exceptions in the Loaded event handler are silently swallowed by WPF.
+                // https://social.msdn.microsoft.com/Forums/vstudio/en-US/9ed3d13d-0b9f-48ac-ae8d-daf0845c9e8f/bug-in-wpf-windowloaded-exception-handling?forum=wpf
+                // http://stackoverflow.com/questions/19140593/wpf-exception-thrown-in-eventhandler-is-swallowed
+                // tl;dr: M$ says it's "by design" and "working as indended" but may change in the future :).
```
objorke: intended

Source/HelixToolkit.Wpf.SharpDX/Controls/DPFCanvas.cs
```diff
@@ -180,15 +180,29 @@ public void InvalidateRender()
...
+
+                // This prevents a crash if no DX10 GPU is present (VMWare)
+                // ToDo: (MVVM friendly) exception handling
```
objorke: log in issues, or do you remember?

Source/HelixToolkit.Wpf.SharpDX/Controls/DPFCanvas.cs
```diff
@@ -180,15 +180,29 @@ public void InvalidateRender()
...
             }
 
-            this.StartD3D();
-            this.StartRendering();
+            try
+            {
+                this.StartD3D();
+                this.StartRendering();
+            }
+            catch (Exception)
```
objorke: should there be a message shown?

smischke:
There are already enough MessageBox.Show() coming up from the EffectsManager when no DX10 GPU present. This is just a quick fix to prevent the crash.

My mid term plan is: If the App registers for and handles error events, don't show anything, else do as now (MessageBox) or maybe display TextBlock with error inside of Viewport3DX. What do you think?

objorke:
I think a TextBlock inside the view would be good! Please no message boxes!

PR #129, opened 2015-03-17
WPF: enable/disable snapping current position to element

chrkon:
We use the HelixToolkit in a WPF application to simplify a given 3D drawing. We mainly create single Points on intersections. This was difficult, because the property "CurrentPosition" of the HelixViewport snaps to the element position if it is in click range around the mouse cursor.
But we don't need this behaviour. We need the real click position. Therefore I have extend the HelixViewPort3D class. I have implemented the new dependency property *SnapMouseToElement*. If this property is set to false the snapping is disabled and we get the coordinates we expect. 

Kind regards,
Christof Konstantinopoulos

### DIFF
--- a/Source/HelixToolkit.Wpf/Controls/HelixViewport3D.cs
+++ b/Source/HelixToolkit.Wpf/Controls/HelixViewport3D.cs
@@ -280,6 +280,13 @@ namespace HelixToolkit.Wpf
                 new UIPropertyMetadata(false, HeadlightChanged));
 
         /// <summary>
+        /// Identifies the <see cref="SnapMouseToElement"/> dependency property.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


objorke:
Hi Christof, this is a good idea! But I would like to see a slightly different implementation... See #133

objorke:
Please include an example for new features! :-) Otherwise I have to do it to test that the code really works. For this feature I would create a new demo "MousePosition" in the Source\Examples\WPF\ExampleBrowser\Examples that simply binds a `TextBlock` to the current mouse position property/ies. Add a cube to the scene to verify that nearest model hit is enabled/disabled.

And please update CONTRIBUTORS/AUTHORS if you are not already listed. :-)

chrkon:
I have worked on #133 so this pull request is obsolete.

Issue #133, opened 2015-03-23
Current mouse position of HelixViewport3D


objorke:
Add functionality to calculate the current 3D mouse position in different ways:

- position of nearest model hit
- intersection with a given plane
- intersection with plane through camera 'target point' (the current behaviour when there is no hit, but maybe this can be removed?)
- 3D ray

objorke:
I suggest to mark `CurrentPosition` and `EnableCurrentPosition` obsolete (but keep the current implementation) and create new properties:
- `CalculateMousePositions` - enable/disable mouse position calculations
- `MouseModelHitPosition` - the position of the nearest hit (or `null` if no intersection)
- `MousePlane` - a `Plane3D` that defines the intersection plane used for mouse position 
- `MousePlaneIntersection` - the position on the `MousePlane` (or `null` if no intersection)
- `MousePosition` calculated as `MouseModelHitPosition ?? MousePlaneIntersection`

objorke:
Any better ideas? Could also create an enum that determines which method to use (none, model hit, plane intersection)

objorke:
Could also use "pointer" instead of "mouse" in the property names, so this feature can also be used for touch events

chrkon:
I have talked with my CEO and I have the permission to work on this issue in my working time.
So I start now on this issue.

Kind regards,
Christof

objorke:
Great! Please squash the commit history and use english in the commit description before doing a PR! :-)

objorke:
Issue taken by @chrkon!

chrkon:
actual changes:

Obsolete Properties| 
-------------------------
`EnableCurrentPosition`| 
`CurrentPosition`| 

New Properties |  Description
-------------------| ----------------
`CalculateCursorPosition`| Enable or disable the calculation of all cursor position properties.
`Cursor3DPosition`| The 3D position based on the target point of the camera. Returns `null`if position could not be calculated.
`CursorOnElementPosition`| The position on a hit visual 3D element. Returns `null`if position could not be calculated or no element is hit.

created but still inoperabel Properties| Description
-----------------| -------------
`CursorRay`| A ray from the screen position to the camera Target point (in case of perspective Camera) or the ray from screen position along the camera LookTo vector (in case of orthographic Camera)
`CursorPlane`| A virtual plane (a construction plane) which can be defined independently from the camera.
[...]

PR #135, opened 2015-03-26
#133 new Cursor Property in HelixViewport3D

chrkon:
New CursorPosition properties are added to the HelixViewport3D.

A `ConstructionPlane` Property is added, too. It holds a simple Plane3D object. This plane is independent from the View, and can be set freely. Per default it is set to 0;0;1 (XY plane).

With these Properties we can now get the following Positions:
* `CursorPosition` ; the 3D Position based on the Camera View 
* `CursorOnElementPosition`; the 3D coordinate on a hit Element. If no Element is hit, this Position is `null`
* `CursorOnConstructionPlane` ; the 3D coordinate on the Construction plane.


chrkon:
Sorry, I was not able to remove the GIT commits with the german text. :-(

objorke:
Looks very good!
1. please do not use [#region](http://programmers.stackexchange.com/questions/53086/are-regions-an-antipattern-or-code-smell)
2. can you please make a new branch on your fork, and squash all the commits into one? this will make the commit history on the master much cleaner

objorke:
What about pulling the latest from HelixToolkit upstream/master, creating a new branch on your fork (e.g. "features/cursor-position") then soft reset (keeping all changes), and create one commit including #133 in the description, create a new PR for the new branch (include #133 in the PR description, not the title)

Sorry for the nitpicking here...

I tried it locally in SourceTree on your fork:
image
But I would like you to do this to get the right Author name! :-)

chrkon:
I try again :-)

PR #143, opened 2015-04-02
SharpDX: Fixed touch to mouse promotion.

smischke:
Due to the CameraController using Manipulation, there is no automatic touch to mouse promotion (see http://nui.joshland.org/2010/04/why-wont-wpf-controls-work-with-touch.html). Thus Viewport3DX doesn't do hit testing and isn't firing Mouse3D events, which makes touch selection impossible. 

Here is a fix. You can see the results in ExampleBrowser's Workitem10048 (you will need a touch device or some way to simulate touch events like the Surface 2.0 SDK).

@przem321 Do we still need the Viewport3DX.Emulate* methods?

### DIFF
--- a/Source/HelixToolkit.Wpf.SharpDX/Controls/Viewport3DX.cs
+++ b/Source/HelixToolkit.Wpf.SharpDX/Controls/Viewport3DX.cs
@@ -229,6 +229,11 @@ namespace HelixToolkit.Wpf.SharpDX
         private Adorner targetAdorner;
 
         /// <summary>
+        /// The <see cref="TouchDevice"/> of the first TouchDown.
+        /// </summary>
+        private TouchDevice touchDownDevice;
+
+        /// <summary>
         /// The view cube.
         /// </summary>
         private ViewCubeVisual3D viewCube;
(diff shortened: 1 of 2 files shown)


Source/HelixToolkit.Wpf.SharpDX/Controls/Viewport3DX.cs
```diff
@@ -1001,69 +1006,92 @@ protected virtual void OnCameraChanged()
...
+            flags |= (int)Mouse.LeftButton << 0;
+            flags |= (int)Mouse.MiddleButton << 1;
+            flags |= (int)Mouse.RightButton << 2;
+            flags |= (int)Mouse.XButton1 << 3;
+            flags |= (int)Mouse.XButton2 << 4;
+            return (MouseButton)flags;
         }
 
-        public void EmulateMouseUpByTouch(MouseButtonEventArgs e)
+        /// <inheritdoc/>
```
objorke: is this supported by VS? or only Sandcastle?

objorke:
@przem321 can you review this?

Source/HelixToolkit.Wpf.SharpDX/Controls/Viewport3DX.cs
objorke: I voted for the feature at https://visualstudio.uservoice.com/forums/121579-visual-studio/suggestions/3745102-add-intellisense-support-for-the-inheritdoc-ta

smischke: It's not supported by VS2013 and at least not in the preview of VS2015. Removing any comment makes intellisense work. I haven't tried if sandcastle also works, but HelixToolkit isn't included in our documentation anyway.

PR #149, opened 2015-04-27
Fixing issue where relative face indices are off by one

garyjohnson:
Having an issue where relative face indices in an OBJ are off by one. Seeing this in OBJ files exported by 3DS Max specifically. When there is a relative index that references the first point in the OBJ file, I end up with an index out of range exception as it ends up resolving to TextureCoordinates[-1] or Normals[-1]. 

Let me know if you have any rules around pull requests and need me to resubmit, I looked around and couldn't find any specific instructions.


objorke:
Great! There are no rules around pull requests yet, I will try to create a checklist soon.
It would be great with an example .obj file with negative indices - maybe this was never tested. I don't remember. It would also be great with a unit test to make sure this bug will not be reintroduced!

garyjohnson:
Definitely would like to include an example OBJ and then test off of that. Unfortunately the models I'm working with are intellectual property and I don't have direct access to 3DS Max to try to reproduce. I will work on figuring out a good way to test this.

objorke:
Writing such an obj "by hand" in a text editor could be an idea. 
I think I have done this before (e.g. https://github.com/helix-toolkit/helix-toolkit/blob/master/Models/obj/Triangle.obj) 

Verify that it imports correctly in other software.

garyjohnson:
Okay, I was finally able to recreate simple models for all of these cases and then implement tests behind them. Tests ensure that absolute and relative points can be parsed from faces and verifies the correct data is parsed.

objorke:
Looks good! Sorry for the late review...

Issue #42, opened 2014-08-27
Portable library (refactoring)


objorke:
Move mesh builder, camera manipulation, exporters and importers to a portable class library (PCL). 

It should be possible to use this on library on different visualization platforms (e.g. WPF, SharpDX, OpenGL).

PR #150, opened 2015-04-27
Fixing SetSmoothingGroup trying to parse a long as an int

garyjohnson:
This was already fixed in SharpDX, just updating ObjReader in WPF to not parse a long value as an int. Having this code in two places adds risk and may be worth trying to share between projects.


objorke:
Totally agree, I would like to refactor the obj import code into the portable library (#42).

PR #152, opened 2015-04-27
Add support for ObjReader to parse line continuations

garyjohnson:
While not explicitly called out in [the specification](http://www.cs.utah.edu/~boulos/cs3505/obj_spec.pdf), the spec is littered with examples containing lines wrapped using a '\' character at the end of the line. A single line can be wrapped multiple times. Specifically Rhino 3D has an obj option to wrap long lines like this. I have no idea why this is useful but it's something I've had to deal with.

I added a couple example OBJs and started a folder (Models\obj\obj_format_tests) for example objs that demonstrate specific features in the OBJ specification for future testing. This feature has been added in both the WPF and WPF.SharpDX ObjReader.

Edit: looks like the terminology is 'line continuation'


JeremyAnsel:
From http://www.fileformat.info/format/wavefrontobj/egff.htm#WAVEOBJ-DMYID.2 :

> OBJ files do not require any sort of header, although it is common to begin the file with a comment line of some kind. Comment lines begin with a hash mark (#). Blank space and blank lines can be freely added to the file to aid in formatting and readability. Each non-blank line begins with a keyword and may be followed on the same line with the data for that keyword. Lines are read and processed until the end of the file. Lines can be logically joined with the line continuation character ( \ ) at the end of a line.

- Does `\` apply to comment lines?
- Are blank lines ignored?

garyjohnson:
Good questions. I searched for a bit on documentation for answers to these and couldn't find any so I did some experimentation with Rhino 3D and Blender on Mac.

Here's what I found:

**Attempt 1:** Load as normal
```
# www.blender.org
v -0.055876 2.197768 0.659192
```
Used a vertex data and comment (in obj of simple sphere) as subjects of test
Blender and Rhino import without issue

**Attempt 2:** Truncate vertex data
```
v -0.055876 2.197768
```
Done to ensure that import would actually fail if vertex data was missing
Blender and Rhino both raise errors and do not import successfully

**Attempt 3:** Blank line continuations
```
v -0.055876 \
\
2.197768 \ 
\
 0.659192
```
Blender and Rhino import without issue

**Attempt 4:** Line continuation on comment
```
# \
\
www.blender\
.org
```
Done to ensure that line continuations work in comments and do not need additional # and do not get parsed as commands
Blender and Rhino import without issue

This suggests that line continuations always merge with the next line regardless of context or content, which I believe is how it is implemented in this PR. I can add some unit tests for these specific cases, however. Let me know how you want to proceed.

JeremyAnsel:
The spec says that blank lines and comment can appear anywhere. I think that it suggests they can appear between merged lines.

I've tested that in Blender:

- line continuation:
OK
```
v 0 0 0
v 7 0 0
v 7 0 7

f \
1 \
2 \
3
```

- line continuation with blank lines:
OK
```
v 0 0 0
v 7 0 0
v 7 0 7

f \

1 \

2 \

3
```

- line continuation char at the end of a comment:
OK
```
v 0 0 0
v 7 0 0
v 7 0 7

# \
f 1 2 3
```

- line continuation with comment:
error
```
v 0 0 0
v 7 0 0
v 7 0 7

f \
#
1 \
2 \
3
```

Source/HelixToolkit.Wpf.SharpDX/Utilities/ObjReader.cs
```diff
@@ -184,6 +184,12 @@ public Object3DGroup Read(Stream s, ModelInfo info = default(ModelInfo))
                         break;
                     }
 
+                    while (line.EndsWith("\\"))
```
JeremyAnsel: We should ignore blank lines.

Source/HelixToolkit.Wpf/Importers/ObjReader.cs
```diff
@@ -194,6 +194,12 @@ public override Model3DGroup Read(Stream s)
                         break;
                     }
 
+                    while (line.EndsWith("\\"))
```
JeremyAnsel: We should ignore blank lines.

Source/HelixToolkit.Wpf.SharpDX/Utilities/ObjReader.cs
garyjohnson: Sorry, could you elaborate on that a little? The while condition will evaluate to false if it's an empty or whitespace string unless I'm missing something.

objorke:
The change to the code looks ok to me. I agree with with @JeremyAnsel that blank lines should also be supported - but this could be another issue/PR. 
I think the example files are too long - I would prefer simple, "hand-written" files like Jeremy shows. Also, I don't think mtl files are needed to test this feature. 
Can you please reduce the test files to the minimum? I think they can all be placed in the models/obj folder, no need to create a sub-folder.

garyjohnson:
Okay, did the following:
* added behavior for dealing with blank lines in the middle of line continuations
* added tests for other line continuation behavior that we figured out in this thread (implicitly worked but wasn't tested before)
* collapsed examples into smaller files without MTLs that better describe behavior
* fixed some naming and some K&R bracket style that crept into tests

garyjohnson:
Also just an aside -- the blank lines in middle of continuation thing works in blender but doesn't in a couple other apps I tried. I don't think it's a big deal and blender is probably a good yardstick anyways, but wanted to mention it.

JeremyAnsel:
The size of the test files could be reduced (instead of a cube, a simple triangle is enough).
Except that, LGTM.

garyjohnson:
Okay I reduced the models to a simple triangle. I also added a test for the model without any continuations as a control test. Finally, during this I noticed that my assertions weren't very robust so I added explicit checks for the vertices and helper methods to keep that junk out of the tests.

objorke:
Looks good! Sorry for the late review...

objorke:
Seems there are some conflicts that must be resolved...

garyjohnson:
Whoops. I'll get this resolved later today and update the PR.

objorke:
thanks, I could have done it - but I really appreciate if you can handle this resolve!

garyjohnson:
Resubmitted as [PR #157](https://github.com/helix-toolkit/helix-toolkit/pull/157) due to egregious user error.

Issue #28, opened 2014-08-27
Make ModelImporter thread safe


objorke:
**This issue was imported from [CodePlex](http://helixtoolkit.codeplex.com/workitem/9997)**

**[pfresnay](http://www.codeplex.com/site/users/view/pfresnay)** wrote 2013-09-20 at 18:59
Some models are heavy to load, ModelImporter should be thread safe to allow background loading without freezing UI.

Trying to load a model in a background thread throw an InvalidOperationException due to not freezed WPF 3D objects.

**[objo](http://www.codeplex.com/site/users/view/objo)** wrote 2013-09-25 at 09:06
Did you try freezing the model (including all geometries etc) after loading and before adding it to the view?

Should the elements always be created on the UI thread?

Can you make a small background loading example that fails.

**[pfresnay](http://www.codeplex.com/site/users/view/pfresnay)** wrote 2013-09-25 at 10:20
@objo thanks for feedback. 

See attached VS2012 project for example. 

I looked at ModelImporter code, it seems that some elements are created using default elements that were previously created in UI thread without being freezed.

PR #158, opened 2015-05-19
Allow model to be loaded on different thread

garyjohnson:
It was important for me to be able to load models on non-UI threads. The main blocker on that was Material / PhongMaterial being DependencyObjects and having thread affinity. I believe I was able to accomplish the same thing by implementing INotifyPropertyChanged in those classes instead. I couldn't find anything that specifically relied on Material being a DependencyObject.

With this change I am now able to run ObjReader.Read on another thread and dispatch the results back to the UI thread. Excuse the ugly code:

```
ThreadPool.QueueUserWorkItem(modelPath => {
    var objReader = new ObjReader();
    var loadedObjects = objReader.Read((string)modelPath);

    Application.Current.Dispatcher.BeginInvoke(DispatcherPriority.Normal, new Action<object> (object3Ds => {
        foreach (var obj3D in (IEnumerable<Object3D>) object3Ds) {
            var mesh = new MeshGeometryModel3D {
                Name = obj3D.Name,
                Geometry = obj3D.Geometry,
                Material = obj3D.Material,
                Transform = ConvertMatrixTransform(obj3D.Transform)
            };
            Models.Add(mesh);
        }
    }), loadedObjects);

}, @"C:\test.obj");
```

### DIFF
--- a/Source/HelixToolkit.Wpf.SharpDX/Model/Materials/Material.cs
+++ b/Source/HelixToolkit.Wpf.SharpDX/Model/Materials/Material.cs
@@ -6,7 +6,6 @@
 
 using System;
 using System.ComponentModel;
-using System.Runtime.CompilerServices;
 
 namespace HelixToolkit.Wpf.SharpDX
 {
@@ -21,7 +20,7 @@ namespace HelixToolkit.Wpf.SharpDX
             set
             {
                 _name = value;
-                OnPropertyChanged();
+                OnPropertyChanged("Name");
             }
         }
 
@@ -31,7 +30,7 @@ namespace HelixToolkit.Wpf.SharpDX
         }
 
         public event PropertyChangedEventHandler PropertyChanged;
-        protected virtual void OnPropertyChanged([CallerMemberName] string propertyName = null) {
+        protected virtual void OnPropertyChanged(string propertyName) {
             var handler = PropertyChanged;
             if (handler != null) handler(this, new PropertyChangedEventArgs(propertyName));
         }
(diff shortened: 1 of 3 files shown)


garyjohnson:
I think this helps pave the way for a `Task<List<Object3D>> ModelReader.ReadAsync(string path)`, but I'd like to spend some time doing this in local code and ensure I don't run into any snags first.

garyjohnson:
Removed CallerMemberName for determining property name to preserve .NET 3.5 compatibility.

garyjohnson:
Found that BitmapImage could also get instantiated for textures / bump maps. Freezing to allow access across threads.

objorke:
Isn't this a breaking change where the materials will not be XAML friendly any more?
Can't this be solved in other ways? Freeze the materials?

I would like to see most of the obj import code moved to the portable library, leaving only a small part in the WPF library - but that is a bigger task... #42

Is this issue related to/duplicate of #28?

@smischke @ikeough what do you think about this change?

garyjohnson:
I hadn't considered this case. 
Which means there would need to be a material class that is a DependencyObject and matching material class that is INotifyPropertyChanged. In which case, one of those use cases would be a breaking change depending how you were accessing / creating materials. 

They could potentially be same name / different namespaces (i.e. `HelixTookit.Wpf.SharpDX.Model.Materials.Material` vs `HelixTookit.Wpf.SharpDX.Model.Materials.Xaml.Material`) so that one of the use cases would potentially only need to change an import.

What are your thoughts? Is there a better alternative?

objorke:
I think it will be quite confusing to have two different versions of materials (and other objects)... :-)
Isn't there an alternative to this. In the similar WPF model (based on dependency objects) I have not seen the need for this!

garyjohnson:
Agreed. I will investigate and see if there's another way around this.

objorke:
Please reopen or create a new PR if you find a good solution to this!
I do not want to merge this PR as it is.

Issue #110, opened 2015-01-15
.obj load  missing texturecoordinates


Drumstig:
hi, im trying to load OBJ files but there seems to be a problem with the texture coordinates

while testing with a simple plane that has 4 vertex points but 6 texturecoordinates i noticed that the returned mesh in the model3dgroup only contains 4 textureCoordinates

i noticed that in objread.cs line 585, it only adds a texturecoordinate when a vertex is added (but 1 vertex can have multiple texturecoordinates when it is part of multiple faces)
(simple unwrap example: http://i.gyazo.com/d84926ad77fd9655e6f9d8e677340ce0.png)

does anyone have an idea on how to solve this problem?

thx in advance!

HIMISOCOOL:
if it's a square texture it will stretch over the whole model normally if you go 
Material = PhongMaterials.DefaultVRML;
Material.DiffuseMap = new BitmapImage(new System.Uri(@"./texturelocation/texture.jpg", System.UriKind.RelativeOrAbsolute));

JeremyAnsel:
A workaround is to disable smoothing groups:

    s off

PR #159, opened 2015-05-19
Fix .obj import issue related to texture coordinates

JeremyAnsel:
This will test and fix the issue reported at #110 .


objorke:
Thanks for the contribution!!

PR #167, opened 2015-06-14
Use $version$ in nuspec files

JeremyAnsel:
Fix #125

PR #170, opened 2015-06-18
Correct version in nuspec files

JeremyAnsel:
Fix #167


JeremyAnsel:
Related to objorke/lynx-toolkit#2

objorke:
Thanks!

PR #172, opened 2015-07-09
Helix Updates from the Dynamo Team

ikeough:
### What It is
This PR adds several new capabilities to `HelixToolkit.Wpf.SharpDX` that were developed by the [Dynamo](https://github.com/DynamoDS/Dynamo) team.

### What It Does
- Transparency sorting for `HelixToolkit.Wpf.SharpDX1`. `MaterialGeometry3D` objects now have a `HasTransparency` flag. When set to `True` these objects will participate in sorting by distance from the camera prior to rendering. Doing this enables good (but not perfect) alpha blending for overlapping transparent objects.
- `GeometryModel3D` objects now have an `IsSelected` property. This property can be used to indicate that a geometry is selected.
- Adds an event, `IntializingEffects` in`Effects.cs` which allows third-party consumers of `Helix.Wpf.SharpDX` to handle the event and register their own effects. In order to make this work, one overload of `RegisterEffect` and one overload of `RegisterLayout` were made public so that third-parties handling the event would have methods available to register their effects.
- Fixes an issue with the re-scaling of the render texture. Previously, when re-sizing the viewport, the view would flash. The view no longer flashes, but the texture stretches until the dispatcher has spare cycles and re-draw can occur. The texture stretch is not a great affect either, but it beats flashing.

### Note
- This PR is still marked as DO NOT MERGE until we can fully separate the Dynamo assets. The last bit of separation that needs to occur is to remove `Dynamo.fx` and all references to the same from `Default.fx`

### PTAL
@objorke @smischke


Source/HelixToolkit.Wpf.SharpDX/Shaders/Default.fx
```diff
@@ -827,3 +816,5 @@ technique11 RenderCubeMap
         SetPixelShader		( CompileShader( ps_4_0, PShaderCubeMap() ) );
     }    
 }
+
+#include "./Shaders/Dynamo.fx"
```
ikeough: This include will be removed when `Dynamo.fx` is removed.

Source/HelixToolkit.Wpf.SharpDX/Shaders/Lines.fx
```diff
@@ -123,12 +124,23 @@ GSInputLS VShaderLines( VSInputLS input )
 		};		
 		output.p = mul( mInstance, output.p );
 	}
-	
+
+	bool isSelected = input.parameters.x;
+	float4 finalColor;
+
+	if (isSelected){
```
ikeough: The line shader is edited to blend the selection color based on `isSelected`.

Source/HelixToolkit.Wpf.SharpDX/Shaders/Material.fx
```diff
@@ -25,8 +25,10 @@
 	bool   bHasInstances		= false;
 	bool   bHasShadowMap		= false;
 
-	float2 vShadowMapSize		= float2(1024, 1024);
-	float4 vShadowMapInfo		= float4(0.005, 1.0, 0.5, 0.0);
+	float2 vShadowMapSize = float2(1024, 1024);
+	float4 vShadowMapInfo = float4(0.005, 1.0, 0.5, 0.0);
+
+	float4 vSelectionColor = float4(0.0, 1.0, 1.0, 1.0);
```
ikeough: `vSelectionColor` is added to `MaterialFx` to be used for coloring objects by selection.

Source/HelixToolkit.Wpf.SharpDX/Shaders/Points.fx
```diff
@@ -54,11 +55,20 @@ GSInputPS VShaderPoints( GSInputPS input )
...
 
+	bool isSelected = input.parameters.x;
+
     //set position into clip space	
     output.p = mul( output.p, mWorld );		
     output.p = mul( output.p, mView );    
     output.p = mul( output.p, mProjection );	
-    output.c = input.c;
+
+	if (isSelected){
```
ikeough: Points can be colored according to selection.

Source/HelixToolkit.Wpf.SharpDX/Model/Elements3D/PointGeometryModel3D.cs
```diff
@@ -314,35 +314,33 @@ public override void Dispose()
...
-                        Color = color,
-                    };
+                    finalColor = color;
                 }
+
+                result[i] = new PointsVertex
+                {
+                    Position = new Vector4(positions[i], 1f),
+                    Color = finalColor,
+                    Parameters = new Vector4(IsSelected?1:0,0,0,0)
```
ikeough: The `PointsVertex.Parameters` field now stores a vector for conveying extra data to the shader. Currently, we only use one component of this vector to store the `IsSelected` value.

Source/HelixToolkit.Wpf.SharpDX/Controls/RenderContext.cs
```diff
@@ -35,6 +35,8 @@ public class RenderContext : IDisposable
 
         public Matrix ProjectrionMatrix { get { return this.projectionMatrix; } }
 
+        public Matrix WorldMatrix { get { return worldMatrix;} }
```
ikeough: This property is created so that third parties which create their own types which inherit from `GeometryModel3D` can overload the `Render` method.

Source/HelixToolkit.Wpf.SharpDX/Controls/ViewportExtensions.cs
```diff
@@ -355,12 +355,20 @@ public static Ray UnProject(this Viewport3DX viewport, Vector2 point2d)//, out V
 
                 var vp = GetScreenViewProjectionMatrix(viewport);
                 var vpi = Matrix.Invert(vp);
+
+                var test = 1f / ((p.X * vpi.M14) + (p.Y * vpi.M24) + (p.Z * vpi.M34) + vpi.M44);
```
ikeough: This check was added to fix an instability in Pan operations when the far clip plane was very large.

Source/HelixToolkit.Wpf.SharpDX/Model/Elements3D/LineGeometryModel3D.cs
```diff
@@ -426,34 +424,30 @@ public override void Dispose()
...
-                        Color = color,
-                    };
+                    finalColor = color;
                 }
+
+                result[i] = new LinesVertex
+                {
+                    Position = new Vector4(positions[i], 1f),
+                    Color = finalColor,
+                    Parameters = new Vector4(IsSelected?1:0,0,0,0)
```
ikeough: `LinesVertex.Parameters` added to hold extra parameters for rendering. See the notes for `PointsVertex` below.

objorke:
Wow - there is a lot of great improvements here!

I like the simple transparency solution, but I am a bit concerned about adding the `HasTransparency` property to the model, since transparency also can be solved by other techniques. Maybe this property is not needed for other techniques? Then it can lead to confusion... 
Can transparency be handled in a similar way as in the WPF solution? https://github.com/helix-toolkit/helix-toolkit/blob/master/Source/HelixToolkit.Wpf/Visual3Ds/Composite/SortingVisual3D.cs
Could the `HasTransparency` property be an attached dependency property?
Does the transparency solution work with visual trees / hierarchies? 

I am also a bit concerned about adding `IsSelected` to the model. I see it can be convenient, but I think the selection could also be kept in a collection (hash set) outside the model, or use an attached dependency property.

To make review easier, it would be great to have one PR for each feature :-) There should also be references from the PR to issues, and the change log should be updated.

Do you know why the build is failing?

ikeough:
@objorke Great feedback. Thanks. Yes, I figured you might balk at having `IsSelected` and `HasTransparency` to `GeometryModel3D`. To be honest, I didn't know about attached dependency properties. I've just been reading up and they seem quite cool. I think the best thing to do is move these properties out to only my object types which inherit from `GeometryModel3D`. I'll remove the `IsSelected` logic in the shaders as well and define my own versions of the point and line shaders.

That will lighten this PR considerably, but there is still the issue of transparency sorting which has to happen at render. I'll look at the example that you provided and try to come up with a better solution.

objorke:
Great, I am looking forward to an update! :-)

I think it is an advantage to put the sorted rendering into a composite model object and not into the Viewport3DX.Render. Only the children of the composite model object need to be sorted. Isn't this less intrusive?

Please update the SharpDX package. Make it a separate PR, then it will be easy to accept :-)

ikeough:
Closed for now. Will update soon!

PR #173, opened 2015-07-09
Update ExtrudedVisual3D.cs

lcouz:
Method "Tessellate"  : Path property must contains at least 2 values, because line 243 : var forward = this.Path[1] - this.Path[0];
Without this, the following code causes an exception :
TubeVisual3D tb = new TubeVisual3D ();
tb.Path = new Point3DCollection();
tb.Path.Add();


objorke:
The change log and contributors files should also be updated :-)

Issue #175, opened 2015-07-24
TubeVisual3D Example doesn't work


interdrift:
I was trying to use the TubeVisual3D example but that didn't work.

I literally copy paste the code and change the namespaces to match my project name. The result is  an ArgumentOutOfRangeException thrown from ExtrudedVisual3D.

The problem arises when the TubeVisual3D is being constructed it executes the constructor of ExtrudedVisual3D making the Path property non null but still empty.

Here's where the error happens : 

https://github.com/helix-toolkit/helix-toolkit/blob/master/Source/HelixToolkit.Wpf/Visual3Ds/MeshVisuals/ExtrudedVisual3D.cs#L243

Here's a link to the example 

https://github.com/helix-toolkit/helix-toolkit/tree/master/Source/Examples/WPF/ExampleBrowser/Examples/Tube

interdrift:
Can someone give a hand here?

ando23:
I didn't have time yet to test it, but the code seems to be weird:

`if (this.Path == null || this.Path.Count >= 2) { return null; }`
versus
`var forward = this.Path[1] - this.Path[0];`, meaning that Path,Count **must** be at least two.

So, my guess would be that the if should be changed to 
`if (this.Path == null || this.Path.Count < 2) { return null; }`

If I have time, I will test it this evening.

ando23:
As suspected, the comparison is inverted. Changing it to `<` seems to fix the immediate problem.

interdrift:
That seems to solve the problem. I assume you will make a PR?

ando23:
I will as soon as I find out how. (I'm new to git.)

interdrift:
:clap:

PR #177, opened 2015-08-05
Fix issue #175

ando23:
The first two items of `this.Path` are referenced and required. The if-condition needs to be inverted.

(First Pull Request ever, please be nice)


objorke:
Thanks! :+1: 
I forgot that the change log and contributors files should also be updated! :-)

Issue #174, opened 2015-07-10
CameraHelper.ZoomExtends "destroys" camera on zero-sized viewport


ando23:
In my use case, the Helix WPF control is created and updated before being shown. This leads to the situation that the size of the viewport is still set to ActualWidth == 0 and ActualHeight == 0.

Calling the CamerHelper.ZoomExtends (via HelixViewPort3D.ZoomExtends) then calculates a vertical field-of-view of NaN (by dividing by zero) and consequenty calls LookAt with an undefined newLookDirection. This results in the camera becoming invalid (NaN position and NaN direction).
![zoomextends](https://cloud.githubusercontent.com/assets/13274270/8619418/33ed0668-2717-11e5-8ebe-af761edfe742.png)

Once the control is shown, the scene is "empty" until the camera is explicitely reset.

File Helpers\CameraHelper.cs, lines 719-722, Revision

ando23:
I made a workaround, please see my branch.

PR #178, opened 2015-08-05
Fix for issue #174

ando23:
On a viewport with height=0 and width=0, the `ZoomExtents` function can result in a camera with invalid postion and view direction.

This fix (rather workaround) does a check and disregards the invalid aspect ratio.


objorke:
Thanks!

PR #184, opened 2015-08-17
Update SharpDX packages to 2.6.3

ikeough:
This PR does the following:

- It updates the SharpDX packages from 2.5.0 -> 2.6.3.
- It remove lines from the `CanvasMock` that were incompatible with the new version of SharpDX, but were not being used anywhere.
- It adds links to the obj model test files in the .net 4.0 project so that the model files are correctly copied to the output directory and the tests pass when compiling that solution. These links already existed in the .net 4.5 project.


ikeough:
@objorke Not sure what the appveyor failure is. Looks like it's trying to install StyleCop from Choclatey, and it fails. I'm relatively certain this does not have to do with this commit.

ikeough:
This was closed as it inadvertently included commits on my local master. This PR has been renewed in [https://github.com/helix-toolkit/helix-toolkit/pull/186](https://github.com/helix-toolkit/helix-toolkit/pull/186)

PR #203, opened 2015-09-09
Reference SharpDX 2.6.3 in nuspec.

ikeough:
This PR updates the `.nuspec` to require a version of SharpDX >= 2.6.3. We have already updated all relevant projects to reference SharpDX 2.6.3, but now when you attempt to use the Helix3D NuGet package in your Visual Studio project, it still installs SharpDX 2.5.0 by default. 

You can get around this currently by updating the SharpDX packages after they are installed, but it's not entirely clear that this is required.

@objorke Please take a look.


objorke:
Thanks for fixing this!

PR #204, opened 2015-09-09
Effect initialization

ikeough:
This PR makes public methods that can be used by a third party to register effects, associate render techniques, and register layouts. This allows third parties to make effects which do not need to be compiled into shader byte code along with the other helix shaders.

This PR also adds better documentation to the methods mentioned above, and pulls out classes into separate .cs files for clarity.

@objorke Please take a look.


objorke:
This looks good! Thanks for cleaning up and making the registration API public!

PR #208, opened 2015-09-10
Replace NotifyPropertyWeaver with PropertyChanged.Fody

ikeough:
The `HelixToolkit.Wpf.SharpDX` solution would not build in Visual Studio 2015. This is because the sample projects were using `NotifyPropertyWeaver`, which does not support Visual Studio 2015. In researching the cause of this, it became clear that `NotifyPropertyWeaver` had been deprecated, and was replaced by the `PropertyChanged.Fody` NuGet package by the same author. Removing the `NotifyPropertyWeaver` tasks from the sample projects and replacing with `PropertyChanged.Fody` allows those projects to build.

https://github.com/Fody/PropertyChanged

@objorke I tested a number of the demo projects after this change and they seemed to work. I am not sure if this change is acceptable given that it's mostly for me to be able to build Helix in VS2015, but the developer of `NotifyPropertyWeaver` does [recommend making this change](
https://github.com/Fody/PropertyChanged/wiki/ConvertingFromNotifyPropertyWeaver).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -193,5 +193,3 @@ FakesAssemblies/
 # Helix Toolkit specific files
 !Models/obj/
 !Models/obj/**/*.obj
-
-*FodyWeavers.xml
--- a/.gitignore
+++ b/.gitignore
@@ -193,3 +193,5 @@ FakesAssemblies/
 # Helix Toolkit specific files
 !Models/obj/
 !Models/obj/**/*.obj
+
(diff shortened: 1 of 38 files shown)


objorke:
Thanks for fixing this issue!

objorke:
Can you also update the NET40 projects?

ikeough:
Will do.

PR #209, opened 2015-09-10
Techniques and Effects Refactoring

ikeough:
### The Big Idea

A third party should be able to create types which extend the built-in `GeometryModel3D` types.  These types should be able to use custom shaders, without having to touch the Helix tookit source. Previously, this was very difficult for the following reasons:  
- All shaders were compiled to binary and bundled with the `HelixToolkit.Wpf.SharpDX` assembly. This prevented third-parties from altering the default rendering behavior, or the set of `RenderTechnique`s that they wanted to support. Using the new `IRenderTechniqueManager` and `IEffectsManager` interfaces, third-parties can now create their own managers for registering techniques and effects.
2. A third-party could not provide a custom vertex type if they wanted to pass custom vertex data to their shaders. Several modifications to `GeometryModel3D` enable inherited classes to provide custom vertex data. 

### IEffectsManager

The `EffectsManager` is currently a singleton. It creates a number of effects that are used for the Helix3D samples, but might not be needed or wanted by third party developers. This PR does the following:
- We create the `IEffectsManager` interface. This interface is compatible with current usages of the `EffectsManager`, but allows for an implementation where third-parties can load whatever effects they like. 
- We remove the `EffectsManager` singleton and create a property, `IRenderHost.EffectsManager`. 
- All references to the singleton are re-pointed to the render host's `IEffectsManager` instance. 
- We create a `DependencyProperty`, `EffectsManager` on `DPFCanvas` so we can provide our own implementation of `EffectsManager`.

### IRenderTechniquesManager

`Techniques` is currently a static class which contains a list of `RenderTechniques`. This design does not allow for extensibility by third parties. So, I've done the following:

- Created the `IRenderTechniqueManager` interface. This interface exposes a method for registering a render technique with the manager and a Dictionary of `RenderTechniques` keyed by their name.
- Made `const string` fields for all `RenderTechnique` names so we didn't have these names as strings in the code.
- Replaced all references to `Technique.Blah` with lookups in the `RenderTechniqueManager.RenderTechniques` dictionary.
- Made a `DependencyProperty`, 'RenderTechniquesManager` on `DPFCanvas` so we can provide our own implementation of `IRenderTechniquesManager`.
- Updated all demos so the `RenderTechniqueManager` is created in the view model.

### GeometryModel3D

- The `Attach` method is made virtual so that a third-party can override it. This method is responsible for creating the vertex buffers passed to the GPU, so it is critical that a third-party can provide vertex arrays using their own vertex types.

### Custom Shader Demo
A new demo project called, Custom Shader Demo has been created which shows all of the above in action.


ikeough:
Need to fix the tests, and rebase some commits for clarity.

Source/Examples/WPF.SharpDX/InstancingDemo/MainWindow.xaml
```diff
@@ -7,7 +7,7 @@
         mc:Ignorable="d"         
         Title="{Binding Title}" Height="500" Width="800">
     <Window.Resources>
-        <x:StaticExtension Member="hx:Techniques.RenderTechniques" x:Key="RenderTechniques"/>
```
ikeough: Because we no longer have a static `Techniques` class, we use our `TechniquesManager` instance. We use a converter to converter a `KeyValuePair` object representing the items in the `ComboBox` back to a `RenderTechnique`.

Source/Examples/WPF.SharpDX/SimpleDemo/MainViewModel.cs
```diff
@@ -40,15 +40,14 @@ public class MainViewModel : BaseViewModel
         public Color4 DirectionalLightColor { get; private set; }
         public Color4 AmbientLightColor { get; private set; }
 
-
         public MainViewModel()
         {
             // titles
-            this.Title = "Simple Demo";
```
ikeough: Whenver I touch a file that has this many redundant uses of `this`, I remove them.

Source/HelixToolkit.Wpf.SharpDX/Controls/DPFCanvas.cs
```diff
@@ -132,7 +133,26 @@ public IRenderer Renderable
         /// </summary>
         Device IRenderHost.Device
         {
-            get { return this.device; }
+            get { return device; }
+        }
+
+        public static readonly DependencyProperty EffectsManagerProperty =
```
ikeough: We add two new `DependencyProperty` on `DPFCanvas`. This allows a third-party to provide an implementation of `IRenderTechniquesManager` and `IEffectsManager`.

objorke:
In the WPF (not SharpDX) I am strictly enforcing StyleCop:
http://stylecop.soyuz5.com/SA1101.html
:-)

Source/HelixToolkit.Wpf.SharpDX/Controls/DeferredRenderer.cs
```diff
@@ -1153,11 +1153,13 @@ public void Dispose()
         /// </summary>
         private class DeferredLightingVariables : IDisposable
         {
-            public DeferredLightingVariables(EffectsManager effectsContainer)
+            public DeferredLightingVariables(EffectsManager effectsContainer, IRenderTechniquesManager techniquesManager)
```
ikeough: For `DeferredLightingVariables` and `ScreenSpaceProcessingVariables`, we inject and `IRenderTechniquesManager` and an `IEffectsManager`.

Source/HelixToolkit.Wpf.SharpDX/Controls/Effects.cs
```diff
@@ -151,42 +142,40 @@ private void InitEffects()
...
-                    Techniques.RenderPositions,
-                    Techniques.RenderNormals,
-                    Techniques.RenderPerturbedNormals,
-                    Techniques.RenderTangents, 
-                    Techniques.RenderTexCoords, 
-                    Techniques.RenderWires, 
-                    Techniques.RenderLines,
-                    Techniques.RenderPoints,
-                    Techniques.RenderBillboard,
+                    renderTechniquesManager.RenderTechniques[DefaultRenderTechniqueNames.Phong],
```
ikeough: `RenderTechinque` lookup is now done on the `RenderTechniques` dictionary, NOT on the static `Techniques` class.

Source/HelixToolkit.Wpf.SharpDX/Controls/Viewport3DX.Properties.cs
```diff
@@ -268,10 +268,10 @@ public partial class Viewport3DX
             "RenderHost", typeof(DPFCanvas), typeof(Viewport3DX), new FrameworkPropertyMetadata(null));
 
         /// <summary>
-        /// The is deferred shading enabled propery
+        /// The Render Technique property
         /// </summary>
         public static readonly DependencyProperty RenderTechniqueProperty = DependencyProperty.Register(
-            "RenderTechnique", typeof(RenderTechnique), typeof(Viewport3DX), new PropertyMetadata(Techniques.RenderPhong, (s, e) => ((Viewport3DX)s).RenderTechniquePropertyChanged()));
+            "RenderTechnique", typeof(RenderTechnique), typeof(Viewport3DX), new PropertyMetadata(null, (s, e) => ((Viewport3DX)s).RenderTechniquePropertyChanged()));
```
ikeough: Because `Techniques` is no longer available as a static class, we can't set the default here. But, the property is set during construction of the view, by binding the `SelectedItem` to the `RenderTechnique` property on the view model. As this property is set when constructing the view model, the combo boxes bound to this selected value "just work".

Source/HelixToolkit.Wpf.SharpDX/Model/Elements3D/PointGeometryModel3D.cs
```diff
@@ -190,19 +190,20 @@ private void OnColorChanged()
         /// <param name="host"></param>
         public override void Attach(IRenderHost host)
         {
-            renderTechnique = Techniques.RenderPoints;
+            renderTechnique = host.RenderTechniquesManager.RenderTechniques[DefaultRenderTechniqueNames.Points];
```
ikeough: All `Model3D` objects now get access to the dictionary of `RenderTechniques` from their `IRenderHost`.

Source/Examples/WPF.SharpDX/SimpleDemo/MainViewModel.cs
objorke: In the WPF (not SharpDX) I am strictly enforcing StyleCop:
http://stylecop.soyuz5.com/SA1101.html
:-)

Source/HelixToolkit.Wpf.SharpDX/Themes/Generic.xaml
```diff
@@ -54,7 +54,7 @@
                                 </hx:CameraController.InputBindings>
                                 -->
                                 <AdornerDecorator x:Name="PART_AdornerLayer" DataContext="{TemplateBinding DataContext}" HorizontalAlignment="Stretch" VerticalAlignment="Stretch">
-                                    <hx:DPFCanvas x:Name="PART_Canvas"/>
+                                    <hx:DPFCanvas x:Name="PART_Canvas" EffectsManager="{Binding EffectsManager}" RenderTechniquesManager="{Binding RenderTechniquesManager}"/>
```
ikeough: The bound values for these new `DependencyProperty` are `IEffectsManager` and `IRenderTechniquesManager` respectively. This allows third-parties to supply their own implementation of both. I was surprised and happy to learn that this data binding still works even when the underlying property specifies an interface.

Source/Examples/WPF.SharpDX/SimpleDemo/MainViewModel.cs
ikeough: @objorke It's your project and I'll play by your rules. But, I would be remiss if I did not recommend you remove this particular rule. By enforcing proper naming of local members, you can make it obvious to the reader of the code that it is a local member, without having to resort to thousands of `this` in the code.

objorke: I think it is OK to continue with no coding style rules on HelixToolkit.Wpf.SharpDX. I don't think it is a problem that the style is not consistent across the projects. I can add a note about this somewhere!

objorke:
This looks like a very good refactoring/enhancement!
Let us know when the PR is ready to merge.
I also guess you want to squash some commits before we merge.

ikeough:
@objorke This one is ready to go. The `CustomShaderDemo` project is an example of this refactoring at work. In that project I demonstrate a completely custom shader workflow with custom types derived from `GeometryModel3D`. Enjoy!

objorke:
This looks good! Thanks for a great improvement!

smischke:
Nice! I'm looking forward to play with it.

smischke:
Ok, I've checked it out. First of all, thanks a lot for your work. Loading shaders at runtime is an absolute must.

There are some things I don't like:
- The removal of the "this."-prefix increases inconsistency, but it can be seen as a matter of taste and I can live with it.
- ExampleBrowser items don't work anymore (see next point).
- The changes will require the Viewport3DX-ViewModel to have at least the properties "EffectsManager" and "RenderTechniquesManager" or it will simply crash. There are no default values which could easily prevent this crash and keep compatibility. The bindings are enforced inside of the default style (generic theme) and they target the DPFCanvas, which I see as more or less an internal element. 

In my opinion, it would be better to:
- Make DPFCanvas.EffectsManager and DPFCanvas.RenderTechniquesManager (also) DependencyProperties of Viewport3DX with default instances as default values (don't enforce bindings)
- Either Bind DPFCanvas.EffectsManager and DPFCanvas.RenderTechniquesManager to Viewport3DX .EffectsManager and Viewport3DX .RenderTechniquesManager in default style (TemplateBinding) or set them in Viewport3DX.OnApplyTemplate.
- DPFCanvas.SupportDeferredRender can be inferred by the active RenderTechnique

I can make a PR if necessary.

PR #216, opened 2015-09-17
Improving and fixing #209

smischke:
This PR is fixing problems identified here:
https://github.com/helix-toolkit/helix-toolkit/pull/209#issuecomment-141070503


objorke:
@ikeough can you review and merge this PR?

smischke:
I had to revert some things to make MVVM exception handling work again. The static instances of DefaultRenderTechniquesManager and DefaultEffectsManager was not a good idea. Now the fallback mechanism is completely in DPFCanvas, so that the exception relaying works out of the box.

PR #218, opened 2015-09-22
SharpDX: DeviceCreationFlags.Debug should not be used in productive mode!

smischke:
See: http://sharpdx.org/forum/4-general/1774-how-to-debug-a-sharpdxexception
See: http://stackoverflow.com/questions/19810462/launching-sharpdx-directx-app-with-devicecreationflags-debug


ikeough:
@smischke This is my bad. I left this in there when debugging the last round of fixes.

Issue #224, opened 2015-11-07
White screen on all samples, Exception when creating Texture in DX11ImageSource


roy-t:
For me all samples result in a white image 

![capture](https://cloud.githubusercontent.com/assets/623634/11014496/1a0277c6-8539-11e5-8d42-28b0c1d73db6.PNG)

 After doing some digging I noticed a `SharpDx.Exception` being thrown and cought in `DX11ImageSource.cs` on line 97:

`this.renderTarget = new Texture(DX11ImageSource.device, target.Description.Width, target.Description.Height, 1, Usage.RenderTarget, format, Pool.Default, ref handle);`

	SharpDX.SharpDXException occurred
	  HResult=-2147024809
	  Message=HRESULT: [0x80070057], Module: [General], ApiCode: [E_INVALIDARG/Invalid Arguments], Message: The parameter is incorrect.

	  Source=SharpDX
	  StackTrace:
	       at SharpDX.Result.CheckError()
	       at SharpDX.Direct3D9.Device.CreateTexture(Int32 width, Int32 height, Int32 levels, Int32 usage, Format format, Pool pool, Texture textureOut, IntPtr sharedHandleRef)
	       at SharpDX.Direct3D9.Texture..ctor(Device device, Int32 width, Int32 height, Int32 levelCount, Usage usage, Format format, Pool pool, IntPtr& sharedHandle)
	       at HelixToolkit.Wpf.SharpDX.DX11ImageSource.SetRenderTargetDX11(Texture2D target) in D:\Projects\helix-toolkit\Source\HelixToolkit.Wpf.SharpDX\Controls\DX11ImageSource.cs:line 97
	  InnerException: 

I've inspected all the variables, here are their values:

`DX11ImageSource.device ={SharpDX.Direct3D9.DeviceEx}`
`target.Decscription.Width = 100;`
`target.Description.Height = 100;`
`format = SharpDX.Direct3D9.Format.A8R8G8B8;`
`handle = {-2147477374};`

The negative handle for the shared resources seems weird to me. But its not always negative while the exception occurs always. All the other values seem perfectly correct. I've tried the samples on another computer, and there everything works perfectly so it seems its a combination of my hardware and the code.

Specifications of the system where it doesn't work
OS: Windows 10
CPU: Intel i5 3540
RAM: 16GB (2x8)
GPU: AMD R9 280x 3GB

The system where all samples dit work is very similar. The biggest difference is that it has an Nvidia card, not an ATI card.

[...]

PR #228, opened 2015-12-09
Fix Windows 10 issue #224

smischke:
On Windows 7 with nVidia Optimus, only one adapter is reported (the nVidia one) and that works regardless of settings. On Windows 10, both adapters (nVidia and Indel) are reported. The one that is active is reported first and the one that is inactive has no outputs and causes the "new Texture()" exception. This PR should fix it.


objorke:
There is a broken test - null path? Maybe something has changed at appveyor, it seems like it has nothing to do with your change...

objorke:
https://github.com/helix-toolkit/helix-toolkit/blob/master/Source/HelixToolkit.Wpf.SharpDX.Tests/Elements3D/MeshGeometryModel3DTests.cs#L36

Issue #229, opened 2015-12-14
"Zoom extents" doesn't work correctly for BillboardText.


ZHZG:
When there is BillboardText, "Zoom extents" doesn't work correctly.

mjwloch:
Looks to me like this bug is still there.. I checked with the current build and I can still reproduce it.

PR #230, opened 2015-12-14
"Zoom extents" doesn't work correctly for BillboardText. #229 

ZHZG:
Don't take BillboardText into consideration when calculating the bounds for zoom extents.

### DIFF
--- a/Source/HelixToolkit.Wpf/Helpers/Visual3DHelper.cs
+++ b/Source/HelixToolkit.Wpf/Helpers/Visual3DHelper.cs
@@ -79,6 +79,12 @@ namespace HelixToolkit.Wpf
             var bounds = Rect3D.Empty;
             foreach (var visual in children)
             {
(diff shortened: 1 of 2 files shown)


Source/HelixToolkit.Wpf/Helpers/Visual3DHelper.cs
```diff
@@ -79,6 +79,12 @@ public static Rect3D FindBounds(this Visual3DCollection children)
             var bounds = Rect3D.Empty;
             foreach (var visual in children)
             {
+                if (visual.GetType() == typeof(BillboardTextGroupVisual3D)
```
objorke: would it be better to check if the visual implements an interface?

objorke: or add a list of types that should be ignored as a parameter?

PR #231, opened 2015-12-15
"Zoom extents" doesn't work correctly for BillboardText. #229

ZHZG:
Don't take BillboardText into consideration when calculating the bounds for zoom extents.

### DIFF
--- a/Source/HelixToolkit.Wpf/Helpers/CameraHelper.cs
+++ b/Source/HelixToolkit.Wpf/Helpers/CameraHelper.cs
@@ -657,7 +657,9 @@ namespace HelixToolkit.Wpf
         /// </param>
         public static void ZoomExtents(this ProjectionCamera camera, Viewport3D viewport, double animationTime = 0)
         {
-            var bounds = Visual3DHelper.FindBounds(viewport.Children);
+            var bounds =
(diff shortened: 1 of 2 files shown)


Source/HelixToolkit.Wpf/Helpers/CameraHelper.cs
```diff
@@ -657,7 +657,9 @@ public static Matrix3D GetInverseTransform(this Camera camera, double aspectRati
         /// </param>
         public static void ZoomExtents(this ProjectionCamera camera, Viewport3D viewport, double animationTime = 0)
         {
-            var bounds = Visual3DHelper.FindBounds(viewport.Children);
+            var bounds =
+                viewport.Children.FindBounds(
+                    new List<Type> { typeof(BillboardTextGroupVisual3D), typeof(BillboardTextVisual3D) });
```
objorke: not open for extension! If you create a new type of screen space visual in your application (not in the Helix Toolkit library), it cannot be ignored...

objorke: so the previous issue I raised has just been moved to this method! :-)

PR #234, opened 2015-12-16
"Zoom extents" doesn't work correctly for BillboardText

ZHZG:
Add an interface that the bounds of the visual will be ignored.
Solves #229

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ All notable changes to this project will be documented in this file.
 - PointSelectionCommand (Wpf)
 - Support subtract mode in CuttingPlaneGroup (Wpf)
 - Handle collection changes for ScreenSpaceVisual3D.Points (Wpf.SharpDX)
+- Ignore visuals that implement IBoundsIgnoredVisual3D in the bounds calculation (#229)
 
 ### Changed
 - XAML namespace prefix changed to `http://helix-toolkit.org/wpf` (Wpf)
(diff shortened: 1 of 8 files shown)


objorke:
It looks good now! Thanks for cleaning up!
Now only two items left:
1. update the change log and contributors file
2. squash the commits (rebase interactively, edit commit message, push -f) Note that you don't need to create a new PR - it is OK to rewrite the commit history here

For the change log, I suggest to add under the "Added" section:
"Ignore visuals that implement IBoundsIgnoredVisual3D in the bounds calculation (#229)"

ZHZG:
I have updated the log, but have difficulty to rebase.

objorke:
OK, good! I will accept this!
Please use rebase in the next PR! :-)

ZHZG:
Hi Øystein,

I have one question about HelixToolkit.

I get the latest HelixToolkit built and find it is based on .Net 4.5 while our Brix project is on .Net4.0.

Much appreciated if you could give some ideas on this.

Thanks,

Zonggui

objorke:
The NuGet package contains both .NET 4.0 and .NET 4.5 binaries (in the lib/NET40 and lib/NET45 folders). NuGet should automatically pick the right one.
The build artifacts contains both NET40 and NET45 binaries.
You can find the latest at
https://ci.appveyor.com/project/objorke/helix-toolkit/build/2015.1.651/artifacts

PR #235, opened 2015-12-17
add default value for MaterialsFile, fix #205

### LINKED ISSUE #205: ObjExporter appears to be broken
This is sample code I'm using to test ObjExporter. I get an exception from the stream that the path parameter is null.  Am I misunderstanding how to use the function?

	var modelImporter = new ModelImporter();
	Model3DGroup modelGroup = new Model3DGroup();
	modelGroup.Children.Add(modelImporter.Load(@"C:\Users\bsee\Downloads\simpleExistingModel.obj"));

	ObjExporter exporter = new ObjExporter();
	exporter.ExportNormals = true;

	//Todo: Why is this showing that the stream path is null?
	using (Stream filestream = File.Create(@"C:\Users\bsee\Downloads\exportedModel.obj"))
	{
		exporter.Export(modelGroup, filestream);
	}

Preffer:
#### Note:
In this patch I assume the `stream` to be a `FileStream`.
#### Reason:
Although the `IExporter.Export` take a `Stream` as the second parameter, but we can see from many places (like `Exporter<T>.FileCreator `) that current implementation has assumed it to be a `FileStream`. So I also take this assumption and leave it to be refactored in the future.

### DIFF
--- a/Source/HelixToolkit.Wpf/Exporters/ObjExporter.cs
+++ b/Source/HelixToolkit.Wpf/Exporters/ObjExporter.cs
@@ -253,6 +253,11 @@ namespace HelixToolkit.Wpf
                 writer.WriteLine("# {0}", this.Comment);
             }
 
+            if (this.MaterialsFile == null)
+            {
+                this.MaterialsFile = Path.ChangeExtension((stream as FileStream).Name, ".mtl");
+            }
+
             writer.WriteLine("mtllib ./" + this.MaterialsFile);
 
             var materialStream = this.FileCreator(this.MaterialsFile);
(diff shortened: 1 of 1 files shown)


objorke:
See comment on #205.
I don´t like the casting to FileStream, then I think it would be better to  specify the filename as a parameter or use FileStream instead of Stream...

Preffer:
I add a check on `stream` that if it is a `FileStream` it will adopt the default value, or else it will remain the old behavior without any 

As for redesign the API interface, I suggest the primary maintainer to do it later instead of change it in a patch.

objorke:
See the comment regarding portability in #205. FileStream is not portable.
I think the current behaviour is good enough - it should be simple to set the MaterialsFile when creating the exporter. See the updated unit tests (#251).

objorke:
Please reopen or create a new issue if it is not possible to use the proposed solution (setting MaterialsFile).

PR #237, opened 2015-12-17
add and enable SwitchYZ

Preffer:
Option `SwitchYZ` is missing in `ObjReader`. It should exist and should be enabled by default to fit the conventions of Wavefront obj format.

What's more, option `SwitchYZ` exists in `ObjExporter` and is enabled by default, which leads to an awkward behavior that model exported by `ObjExporter` is incompatible with `ObjReader`.

### DIFF
--- a/Source/HelixToolkit.Wpf/Importers/ObjReader.cs
+++ b/Source/HelixToolkit.Wpf/Importers/ObjReader.cs
@@ -52,6 +52,7 @@ namespace HelixToolkit.Wpf
             : base(dispatcher)
         {
             this.IgnoreErrors = false;
+            this.SwitchYZ = true;
 
             this.IsSmoothingDefault = true;
             this.SkipTransparencyValues = true;
@@ -80,6 +81,11 @@ namespace HelixToolkit.Wpf
         /// The default value is on (true).
         /// </remarks>
         public bool IgnoreErrors { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to switch Y and Z coordinates.
+        /// </summary>
+        public bool SwitchYZ { get; set; }
 
         /// <summary>
         /// Gets or sets a value indicating whether to skip transparency values in the material files.
@@ -629,7 +635,14 @@ namespace HelixToolkit.Wpf
         private void AddNormal(string values)
         {
             var fields = Split(values);
-            this.Normals.Add(new Vector3D(fields[0], fields[1], fields[2]));
+            if (SwitchYZ)
+            {
+                this.Normals.Add(new Vector3D(fields[0], -fields[2], fields[1]));
+            }
+            else
+            {
+                this.Normals.Add(new Vector3D(fields[0], fields[1], fields[2]));
+            }
         }
 
         /// <summary>
@@ -653,7 +666,14 @@ namespace HelixToolkit.Wpf
         private void AddVertex(string values)
         {
             var fields = Split(values);
-            this.Points.Add(new Point3D(fields[0], fields[1], fields[2]));
+            if (SwitchYZ)
+            {
+                this.Points.Add(new Point3D(fields[0], -fields[2], fields[1]));
+            }
+            else
+            {
+                this.Points.Add(new Point3D(fields[0], fields[1], fields[2]));
+            }
         }
 
         /// <summary>
--- a/Source/HelixToolkit.Wpf/Importers/ObjReader.cs
+++ b/Source/HelixToolkit.Wpf/Importers/ObjReader.cs
@@ -52,7 +52,7 @@ namespace HelixToolkit.Wpf
             : base(dispatcher)
         {
             this.IgnoreErrors = false;
-            this.SwitchYZ = true;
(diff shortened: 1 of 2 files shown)


objorke:
Agree that this should be consistent.
Tests are broken, cannot merge before build is green.
Can you add a test that does an export-import roundtrip?

Preffer:
I has tweaked the test cases, and all of them has passed on my local machine
capture
But it is still failed on appveyor, I will appreciate what I has missed

JeremyAnsel:
The failed tests are those for the `SharpDX` version of `ObjReader`.
`SwitchYZ` is not implemented in `HelixToolkit.Wpf.SharpDX\Utilities\ObjReader.cs`.

Models/obj/obj_format_tests/line_continuation_empty_continuation.obj
```diff
@@ -1,11 +1,7 @@
...
-v \
-    -1.000000 \
-    \
-    0.000000 \
-    1.000000
-v 1.000000 0.000000 1.000000
-v -1.000000 0.000000 -1.000000
+v -1.000000 1.000000 -0.000000
+v 1.000000 1.000000 -0.000000
+v -1.000000 -1.000000 -0.000000
```
JeremyAnsel: The line continuation character `\` need to be preserved.

Models/obj/obj_format_tests/line_continuation_empty_line.obj
```diff
@@ -1,10 +1,7 @@
...
 # www.blender.org
-v -1.000000 \
-    0.000000 \
-
-    1.000000
-v 1.000000 0.000000 1.000000
-v -1.000000 0.000000 -1.000000
+v -1.000000 1.000000 -0.000000
+v 1.000000 1.000000 -0.000000
+v -1.000000 -1.000000 -0.000000
```
JeremyAnsel: The line continuation character `\` and the empty line need to be preserved.

Models/obj/obj_format_tests/line_continuation_multiple_breaks.obj
```diff
@@ -1,9 +1,7 @@
 # Blender v2.74 (sub 0) OBJ File: ''
 # www.blender.org
-v -1.000000 \
-0.000000 \
-1.000000
-v 1.000000 0.000000 1.000000
-v -1.000000 0.000000 -1.000000
+v -1.000000 1.000000 -0.000000
+v 1.000000 1.000000 -0.000000
+v -1.000000 -1.000000 -0.000000
```
JeremyAnsel: The line continuation character `\` need to be preserved.

Models/obj/obj_format_tests/line_continuation_single.obj
```diff
@@ -1,8 +1,7 @@
 # Blender v2.74 (sub 0) OBJ File: ''
 # www.blender.org
-v -1.000000 \
-0.000000 1.000000
-v 1.000000 0.000000 1.000000
-v -1.000000 0.000000 -1.000000
+v -1.000000 1.000000 -0.000000
+v 1.000000 1.000000 -0.000000
+v -1.000000 -1.000000 -0.000000
```
JeremyAnsel: The line continuation character `\` need to be preserved.

Preffer:
@JeremyAnsel Thanks a lot! I have add SwitchYZ in SharpDX version of `ObjReader` and preserve the line continuation character `\`.

Source/HelixToolkit.Wpf/Importers/ObjReader.cs
```diff
@@ -52,6 +52,7 @@ public ObjReader(Dispatcher dispatcher = null)
             : base(dispatcher)
         {
             this.IgnoreErrors = false;
+            this.SwitchYZ = true;
```
objorke: This is a breaking change. I think it would be better to set the default value to `false`, then all the changes to the test files can be reverted!

objorke:
Can you change the default value of the `SwitchYZ` property?
Then revert changes to the tests. 
Remember to update the change log.
Then squash the commits (rebase -i) into a single commit and `git push -f` to update the PR.

Preffer:
But in `ObjExporter` it is set to `true` by default, if the default value is false, it will lead to an awkward behavior that model exported by `ObjExporter` is incompatible with `ObjReader`.

Preffer:
If you consider it's more important to keep backward compatible than to fix this awkward behavior, I will revert them.

objorke:
Sorry I did not check the `ObjExporter`, I see the default is `SwitchYZ = true` there:
https://github.com/helix-toolkit/helix-toolkit/blob/master/Source/HelixToolkit.Wpf/Exporters/ObjExporter.cs#L84

I agree that the default value of `SwitchYZ` should be the same in both classes. 

But I would prefer default value of `false`. If you want Y up, I think you should change your camera, not transform the coordinates...

I can accept this PR, but it would be great to hear some opinions about the default value :-)

Preffer:
OK, set to `false` by default.

PR #243, opened 2016-01-06
use Point3DCollection instead of List<Point3D>

Preffer:
Reasons:

1. Use `Point3DCollection` which implement `IFormattable` enable usage like this in xaml:
`<Helix:LinesVisual3D Points="-100,0,0 100,0,0" />`
2. `Point3DCollection` is the standard data type in WPF. `MeshGeometry3D.Positions` is `Point3DCollection`.
3.  `Point3DCollection` also implement  `IList`, so it just add new functionalities without lost.

### DIFF
--- a/Source/HelixToolkit.Wpf/Visual3Ds/ScreenSpaceVisuals/BoundingBoxWireFrameVisual3D.cs
+++ b/Source/HelixToolkit.Wpf/Visual3Ds/ScreenSpaceVisuals/BoundingBoxWireFrameVisual3D.cs
@@ -53,7 +53,7 @@ namespace HelixToolkit.Wpf
                 return;
             }
 
-            var points = new List<Point3D>();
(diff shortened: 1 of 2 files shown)


objorke:
Thanks, this looks ok.
PS. Please create an issue before making pull requests. Also update the change log and contributors files.

chrkon:
This change is breaking our application. :-(
Therefore I have forked the Helix Project and I have removed the changes made by this commit.

objorke:
@chrkon see #263

chrkon:
@objorke: it works now!

Switching from `ObservableCollection<Point3D>` to `Point3DCollection` was a breaking change for our Application. Even if I changed our internal declarations to `Point3DCollection` the software was not working poroperly.

**But after your changes in #263 it works again.**
Thank you very much.

PR #245, opened 2016-01-13
UWP port

CobraCalle:
I've created a UWP-port from the Win8-Version, but that doen not work at the moment.

But I've had to change the threshold used when normalizing a polygon from 1E-8 to 1E-9 because I habe a geometry in a project that created an exception at that poin. can you please change that threshold so that I can use the "official" nugets again?


Source/Examples/UWP/SimpleDemo/MainPage.xaml
```diff
@@ -0,0 +1,26 @@
...
+    xmlns:helixModel="using:HelixToolkit.UWP.CommonDX"
+    mc:Ignorable="d">
+
+    <Grid Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
+        <Grid.RowDefinitions>
+            <RowDefinition Height="80"/>
+            <RowDefinition/>
+        </Grid.RowDefinitions>
+
+        <TextBlock Text="HelixToolkit + SharpDX in Windows 8" HorizontalAlignment="Center" VerticalAlignment="Center" FontSize="64" FontStretch="Condensed"/>
```
objorke: in Windows Universal

Source/HelixToolkit.Wpf/Geometry/Polygon3D.cs
```diff
@@ -116,7 +116,7 @@ public Vector3D GetNormal()
             for (int i = 2; i < this.Points.Count; i++)
             {
                 var n = Vector3D.CrossProduct(v1, this.Points[i] - this.Points[0]);
-                if (n.LengthSquared > 1e-8)
+                if (n.LengthSquared > 1e-9)
```
objorke: can you create an issue + a separate PR for this one?

objorke: i.e. revert this line in this PR

objorke:
Great!
It builds fine, but I can't see any cube on my Windows 10... Only black background. No error messages.

objorke:
I think `project.lock.json` should be added to .gitignore

CobraCalle:
Yes, I know... UWP port does not work at the moment. At the moment I only need the changed threshold to be "official"... but did not know how to separate that. 

I've chagne the threshold in Polygon3D once again because 1E-9 does not work on all cpu... in some cases the calculated value ist even smaller, so I've change it to 1E-10

CobraCalle:
At the moment I don't know what's the problem with the uwp version... perhaps some one with more SharpDX know how can have a look at it

CobraCalle:
Hmmm... I've no clue how to start another pull request... the Threshold in Polygon3D has to be 1E-10 instead of 1E-9

objorke:
@CobraCalle - you should create a new branch (based on the upstream master) and "cherry pick" the Polygon3D commit. Then you could create a new PR containing only that change.

Added issue #246 regarding the Polygon3D threshold value. 
I cherry picked the last commit in #247.

objorke:
I have merged the Polygon3D commit.
Alt 1: For the UWP branch, you should now rebase it on the latest upstream master. Rebase interactively (delete/squash commits) to get rid of the Polygon3D commits. Then git push -f
Alt 2: You can create a new branch for your UWP commits, and rebase/cherry pick to clean up the commit history. Create a new PR

CobraCalle:
I've ported your version to use a SwapChainPanel and now I've a first working version. Resizing the ViewportX does not work at the Moment, but I shows the cube ;-)

objorke:
Beautiful! Do you want to implement the resizing before we merge?

CobraCalle:
I'm very busy at the moment and don't know when I'll have time to have a look at the resizing thing... perhaps there are other people out there interested in using Helix on UWP... so I think it would be better to merge it now

CobraCalle:
Do you think that we can port more thing from the desktop version over (Camera Control and so on)?

CobraCalle:
One other question... what's the reason for the new Win8 / UWP projects? Why not just port Helix.WPF.SharpDx over to UWP instead of starting at Adam and Eva?

objorke:
Starting small on a new platform is usually a good idea :-)
We should also get a test project up and running before adding more functionality, in my opinion.
Could the code shared with WPF.SharpDX go into a shared project? How much will be common?
I think the Win8 project can be deleted, focus on UWP only...

PR #250, opened 2016-01-25
Solved drag&drop problem: a lot of DragEnter/DragLeave events when dragging over HelixViewport3D and its children.

isukces:
Solved drag&drop problem mentioned http://forum.helix-toolkit.org/topic/509250-drag-drop-behaviour-changing-between-helixtoolkit-and-helixtoolkitwpf/

A lot of DragEnter/DragLeave events when dragging over HelixViewport3D and its children.

### DIFF
--- a/Source/HelixToolkit.Wpf/Controls/HelixViewport3D.cs
+++ b/Source/HelixToolkit.Wpf/Controls/HelixViewport3D.cs
@@ -3224,6 +3224,7 @@ namespace HelixToolkit.Wpf
                 if (this.adornerLayer != null)
                 {
                     this.adornerLayer.Child = this.viewport;
+                    adornerLayer.IsHitTestVisible = false;
                 }
             }
 


Source/HelixToolkit.Wpf/Controls/HelixViewport3D.cs
```diff
@@ -3224,6 +3224,7 @@ public override void OnApplyTemplate()
                 if (this.adornerLayer != null)
                 {
                     this.adornerLayer.Child = this.viewport;
+                    adornerLayer.IsHitTestVisible = false;
```
objorke: no side effects from this change?
I guess the adorner layer never needs hit testing.
Maybe it should be set in xaml instead: https://github.com/helix-toolkit/helix-toolkit/blob/master/Source/HelixToolkit.Wpf/Themes/Generic.xaml#L94

objorke: you forgot "this." - please follow style of existing code ;-)

objorke:
I moved the code into the control template #253.
Does that work?

objorke:
I merged #253. Please create an issue if it does not work.

isukces:
Yes, drag&drop works properly. Thank you.

desgriffiths:
I think this has the side-effect that the manipulator example no longer works.

objorke:
@desgriffiths thanks for the notice! You are right, this PR broke a lot of other examples!

objorke:
I applied a fix

PR #255, opened 2016-02-02
Update StudioReader.cs

SiNeumann:
Use luminance as EmissiveColor in 3ds Importer, if 3ds doesn't contain luminance it falls back to Transparent anyway

### DIFF
--- a/Source/HelixToolkit.Wpf/Importers/StudioReader.cs
+++ b/Source/HelixToolkit.Wpf/Importers/StudioReader.cs
@@ -547,7 +547,7 @@ namespace HelixToolkit.Wpf
                         }
 
                         mg.Children.Add(new SpecularMaterial(new SolidColorBrush(specular), specularPower));
-
+                        mg.Children.Add(new EmissiveMaterial((new SolidColorBrush( luminance))));
                         if (name != null)
                         {
                             this.materials[name] = mg;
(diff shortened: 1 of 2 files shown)


Source/HelixToolkit.Wpf/Importers/StudioReader.cs
```diff
@@ -547,7 +547,7 @@ private void ReadMaterial(BinaryReader reader, int chunkSize)
                         }
 
                         mg.Children.Add(new SpecularMaterial(new SolidColorBrush(specular), specularPower));
-
+                        mg.Children.Add(new EmissiveMaterial((new SolidColorBrush( luminance))));
```
objorke: why the extra () and space? ;-)

objorke:
Do we have an example where we can see how this looks?

Source/HelixToolkit.Wpf/Importers/StudioReader.cs
objorke: could this line be added only if luminance is not transparent/undefined?

SiNeumann: Maybe I was to fast with the pull request, I'll add an if clause. Anyway this was just a quirky workaroun, as I use the HelixToolkit.WPF.StudioImport in the Sharp.DX extension. Basically just converting the resulting WPF MeshGeometry3D into the according SharpDX type. So I should add a proper handling for this either by
1. Adding a proper Converter from WPF MeshGeometry3d to SharpDX 
or
2. Port the StudioImporter to HelixToolkit.Wpf.SharpDX

Sorry for the inconvenience.

objorke: Ok, the functionality of the wpf 3DS importer should not be broken.
The plan is to move the importers and exporters into the portable library #42.
I prefer your alternative 1.

PR #256, opened 2016-02-17
Port of StudioReader to SharpDX

SiNeumann:
Ported line by line StudioReader to SharpDX, this direct approach was used because converting Geometries from Helixtoolkit.Wpf using the Studioreader there would result in some ugly artifacts. TODO Provide Demo. I have a working Demo on my machine still have to clean it

### DIFF
--- a/Source/HelixToolkit.Wpf/Importers/StudioReader.cs
+++ b/Source/HelixToolkit.Wpf/Importers/StudioReader.cs
@@ -547,7 +547,6 @@ namespace HelixToolkit.Wpf
                         }
 
                         mg.Children.Add(new SpecularMaterial(new SolidColorBrush(specular), specularPower));
-                        mg.Children.Add(new EmissiveMaterial((new SolidColorBrush( luminance))));
                         if (name != null)
                         {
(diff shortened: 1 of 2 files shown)


Source/HelixToolkit.Wpf/Importers/StudioReader.cs
```diff
@@ -547,7 +547,7 @@ private void ReadMaterial(BinaryReader reader, int chunkSize)
                         }
 
                         mg.Children.Add(new SpecularMaterial(new SolidColorBrush(specular), specularPower));
-
+                        mg.Children.Add(new EmissiveMaterial((new SolidColorBrush( luminance))));
```
objorke: Is this needed?
Is it a defect or enhancement?
Can we see this effect on any of the included models? (/Models/3ds)

SiNeumann: Sorry forgot to remove it. It's a defect

objorke: Will not merge until it is reverted :-)

PR #260, opened 2016-03-02
"Zoom extents" doesn't work correctly for BillboardText #234

mjwloch:
Added IBoundsIgnore support to all FindBounds overloads


objorke:
#234

Issue #262, opened 2016-03-11
Problem in PointAndLinesBinding example


KKauK:
1. ) Due to switching from ObservableCollection<Point3D> to Point3DCollection example PointAndLinesBinding throws an exception.

2.) Grafic in PointAndLinesBinding example does not refresh after adding new lines. It always requires "zoom in" with mouse wheel to show lines and points

chrkon:
Part 1 (the exception) seems to be solved if line 40 of file "ScreenSpaceVisual3D.cs" is changed from 
` "Points", typeof(IList<Point3D>), typeof(ScreenSpaceVisual3D)`
to
` "Points", typeof(Point3DCollection), typeof(ScreenSpaceVisual3D)`

Unfortunately the graphic will not be shown in the PointAndLineBinding example.

KKauK:
You will also need to change the definitions and the corresponding properties in the MainWindow.xaml.cs file of the PointsAndLineBinding example to Points3DCollection:

private readonly Point3DCollection linePoints = new Point3DCollection();
private readonly Point3DCollection points = new Point3DCollection();
.....

The grafic will show up as soon as you activate Line or Points in the menu of the example AND than scroll the mouse wheel once.

chrkon:
Yes, now I can see the same behavior. 

For me this behaviour seems to be correct, because the `IObservable Interface` is not part of the `Point3DCollection Class`. Because of this, the system is not notified about changes in the Point3DCollection.

Perhaps a `NotifyPropertyChanged` is missing?

holance:
Fixed.

PR #263, opened 2016-03-15
Fix problem in PointAndLinesBinding example 

objorke:
To partially solve #262
I think the event should use a weak event manager


chrkon:
The "PointAndLinesBinding" example seems to be much faster than before. 
- previous Version: 500 Points = 2 FPS, 1000 Points = 0 FPS
- actual Version: 500 Points = 67 FPS, 1000 Points = 67 FPS  (!)
 
Very good!

PR #266, opened 2016-03-24
Fix Obj texture loading with SharpDX

deviloper:
I edited some lines of code to enhance texture loading when they are linked to an MTL file with the Helix SharpDX ObjLoader. 

User won't have to copy textures in a "media" folder but their path will be relative to the OBJ/MTL file folder.

One possible issue is the loading of textures specified with absolute paths (some softwares export obj/mtl in such format) but I think this change would be useful to many users.

### DIFF
--- a/Source/HelixToolkit.Wpf.SharpDX/Utilities/ObjReader.cs
+++ b/Source/HelixToolkit.Wpf.SharpDX/Utilities/ObjReader.cs
@@ -1180,13 +1180,13 @@ namespace HelixToolkit.Wpf.SharpDX
                     //AmbientMap = this.AmbientMap,
 
                     DiffuseColor = this.Diffuse,
-                    DiffuseMap = (this.DiffuseMap == null) ? null : LoadImage(this.DiffuseMap),
+                    DiffuseMap = (this.DiffuseMap == null) ? null : LoadImage(Path.Combine(texturePath, this.DiffuseMap)),
 
                     SpecularColor = this.Specular,
                     SpecularShininess = (float)this.SpecularCoefficient,
                     //SpecularMap = this.SpecularMap,
 
-                    NormalMap = (this.BumpMap == null) ? null : LoadImage(this.BumpMap),
+                    NormalMap = (this.BumpMap == null) ? null : LoadImage(Path.Combine(texturePath, this.BumpMap)),
                     //Dissolved = this.Dissolved,
                     //Illumination = this.Illumination,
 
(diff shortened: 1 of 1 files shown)


Source/HelixToolkit.Wpf.SharpDX/Utilities/ObjReader.cs
```diff
@@ -1180,13 +1180,13 @@ private Material CreateMaterial(string texturePath)
                     //AmbientMap = this.AmbientMap,
 
                     DiffuseColor = this.Diffuse,
-                    DiffuseMap = (this.DiffuseMap == null) ? null : LoadImage(this.DiffuseMap),
+                    DiffuseMap = (this.DiffuseMap == null) ? null : LoadImage(Path.Combine(texturePath, this.DiffuseMap)),
```
objorke: I think it would be better to set the path in line 925, and remove the texturePath from this method.

Source/HelixToolkit.Wpf.SharpDX/Utilities/ObjReader.cs
```diff
@@ -1199,7 +1199,7 @@ private Material CreateMaterial(string texturePath)
 
             private static BitmapImage LoadImage(string path)
             {
-                var bmp = new BitmapImage(new Uri(@"./Media/" + path, UriKind.RelativeOrAbsolute));
+                var bmp = new BitmapImage(new Uri(path,UriKind.RelativeOrAbsolute));
```
objorke: ok!

objorke: (but there is a missing space... :-)

objorke:
Are all examples still working? If not, we should have an example to test the obj loader!

deviloper:
I tried to load complex obj models obtained from a point cloud and it worked.

Source/HelixToolkit.Wpf.SharpDX/Utilities/ObjReader.cs
deviloper: yes, this would be a better solution. How can i update my pull request?

objorke:
I see that none of the included examples that use the ObjReader contain textures. So the existing examples are working. But I cannot test that the new version works with textures, and don't have time to write an example myself right now. Will merge anyway, hopefully everything is good :-)

Source/HelixToolkit.Wpf.SharpDX/Utilities/ObjReader.cs
izrik: @deviloper Just add a commit to your branch and push it to your repo. GitHub will automatically see the new commit and include it in the PR.

PR #305, opened 2016-09-21
Added TypeConverter to Light3D.Attenuation

Fireforge:
Allows Attenuation values to be set in XAML, which is the best place to put them for static lights IMHO. Seems to have been forgotten when new converters were introduced in https://github.com/helix-toolkit/helix-toolkit/commit/eecffd3be16da3449dfe5f39315ad64dccc46368.


objorke:
thanks for the fix!

PR #312, opened 2016-10-10
Some updates

holance:
1. Create buffer directly to avoid redundant copy data to DataStream in RenderUtil.CreateBuffer<T>.
2. Bug fix. Null check for geometry in PointGeometryModel3D.
3. Added more raster state description properties to MeshGeometryModel3D. Use common OnRasterStateChanged to handle raster related property change.  This will allow multiple MeshGeometryModel3D to share same geometry and show its 3D model, its wireframe and inner surface at the same time by using different raster settings.


objorke:
This looks good to me! Thanks for contributing!

PR #313, opened 2016-10-12
Bug fix and example fix

holance:
Geometry data check to prevent create buffer error on zero size positions/indices.
Example fix for OnRasterStateChanged in CustomShaderDemo.
Multiple bug fix.
Fix hit test for billboard.


objorke:
thanks!

Issue #311, opened 2016-10-09
Incorrect search for best multisampling configuration


BetaRavener:
There are multiple reports of antialias problems when using basic renderer, two examples:
- [Stack Overflow](http://stackoverflow.com/questions/39529885/how-to-smoothen-edges-in-helix-toolkit-sharpdx-viewport)
- [Helix Forums](http://forum.helix-toolkit.org/topics/1143-sharpdx-why-does-my-model-look-better-when-using-the-deferred-rendertechnique-instead-of-eg/)

I have also experienced problems with antialiasing - I am sure that both my integrated Intel GPU and mobile NVidia supports it to some extent. I have looked into the code of `DPFCanvas.cs`, where I found section responsible for picking the best SampleDescription. The logic there basically says:
1. Start at 8 samples
2. Check if sample quality is bigger than 0
2a. True - Use this configuration
2b. False - Lower samples and repeat

However, there are 2 major issues:
- Sample count can be as large as 32
- Sample quality of 0 is still "something". Current code will eventually default into 1:0 (sample:quality), but 8:0 configuration is still looking much better than 1:0.

Compare these 2 images:
8:0 configuration: https://s18.postimg.org/hh95gp8zd/image.png
1:0 configuration: https://s18.postimg.org/njgs76xfd/non_aa.png

So the solution should look something like this:
```
int sampleCount = 1;
int sampleQuality = 0;
if (IsMSAAEnabled)
{
  do
  {
    var newSampleCount = sampleCount * 2;
    var newSampleQuality = device.CheckMultisampleQualityLevels(Format.B8G8R8A8_UNorm, newSampleCount) - 1;

    if (newSampleQuality < 0)
      break;

    sampleCount = newSampleCount;
    sampleQuality = newSampleQuality;
  } while (sampleCount < 32);
}
var sampleDesc = new SampleDescription(sampleCount, sampleQuality);
```

I'm quite surprised this haven't been reported before, either everyone is using powerful GPUs that would give `SampleQuality > 0` or noone is interested in using SharpDx.

**EDIT:** It would be great if user could manually select configuration for SampleDescription, Sometimes selecting the highest possible configuration can be a performance problem, where some lower configuration would suffice.

danZib:
Oh this worked beautifully! Thank you BetaRavener. 
@objorke Any changes this can be included in a future commit, is there some kind of hook?

smischke:
[...]

PR #317, opened 2016-10-22
Fix incorrect search for best multisampling configuration (Issue #311). Credits to @BetaRavener!

smischke:
See issue #311. Credits to @BetaRavener!


objorke:
Great! thanks!!

Issue #180, opened 2015-08-06
TorusVisual3D


juno-craft:
Hi guys, I'm really thankful for this amazing package.

It would be really great if anyone is willing to throw a torus implementation as a Visual3D like this one here ;

http://docs.autodesk.com/ACD/2010/ENU/AutoCAD%202010%20User%20Documentation/index.html?url=WS1a9193826455f5ffa23ce210c4a30acaf-688d.htm,topicNumber=d0e94758

interdrift:
Now that I think more about it this could probably be done by using the PipeVisual3D.

Iluvatar82:
Hey guys,
sorry, I am new to Git, so please be tolerant when I made something wrong here.
I wanted to help a little and created a new branch called **torusvisual3d-branch** via the desktop app. Then I did some coding and committed the changes I made to that branch.

The branch includes a modified MeshBuilder (added a GetCircleSegment function and a AddTorus function), a TorusVisual3D and a new simple Project (in the Example Browser).

I am just now sure how you can now see my additions.

Issue #74, opened 2014-11-06
Add TorusVisual3D


objorke:
See http://helixtoolkit.userecho.com/topic/535265-helixvisual3d-not-updating-after-created/
But implement with a single mesh!
Implement in a similar way as HelixVisual3D

PR #318, opened 2016-10-23
Added a TorusVisual3D and an Example Project to demonstrate it's use (issue #180, #74)

Iluvatar82:
A TorusVisual3D was added. It uses the enhanced MeshBuilder to generate
the Geometry for a Torus. The Torus handles special cases like
self-intersection and restricts the values to a valid range.
A simple demo was also added to the example browser to demonstrate the
usage.
This was written in response to issues #180 and #74


objorke:
this looks good! thanks for contributing!

objorke:
#180 #74

PR #320, opened 2016-10-27
Advanced ViewCube and speedUP of HitTesting in GeometryModel3D

jotschgl:
Added the ability to click on every corner and every Edge on the ViewCube (#1) 

Speed up Hittesting ing the Geometrymodel3D by not using the whole geometry but only the Bounds


objorke:
I agree it would be nice to have a dependency property that enables the edge click feature!

jotschgl:
The text on each side is the same as before. I just added the edges and corners.
But I also think that it would be nice to have the option to dis-/enable the feature.

holance:
Seems like there is a build error due to lack of Summary for DependencyProperties?

objorke:
yes, missing comment on HelixViewport3D.IsViewCubeEdgeClicksEnabled

holance:
Hi Oystein,

The comments have been fixed. #320 is ready to merge.

Best,
Lunci

Issue #327, opened 2016-11-04
(Helixtoolkit.SharpDX)Rotate geometry is sluggish


holance:
If zoom in the geometry very large in viewport and start doing quick mouse rotation, the rotation becomes sluggish every few seconds. GC issue? Race condition?

holance:
Infinite rotate does not have this issue. Seems it happens only using mouse to do fast rotation.

BetaRavener:
I have been looking into this problem before, but couldn't find the cause. At least I think it's the same problem, but instead of "becoming sluggish every few seconds", for me it was "becoming sluggish/unusable after few moments of camera rotation".  I have profiled my aplication with Visual Studio and what I found was that during camera rotation, there was increasing gap between issuing command and graphics card actually executing it. The gap can be though of as lag between input and screen output, and so with longer camera rotation the lag became worse and worse. 

Now, I think part of the issue is that Helix is picking up only integrated Intel HD4000 (I have mobile NVidia but it's in Optimus topology and so helix discards it because it doesn't have screen output). Computing power of that Intel graphics is just bad and it simply can't render scene 60FPS. Still, I would expect the scene to lag "constantly", say 10 FPS, instead of crawling from 60FPS to unusable.

However, one lead I found and it might be important piece to solving this: The way that Helix measures drawing time in SharpDx part is just plain wrong. It uses CPU timer to time how much rendering of the frame took BUT what it really does is just measuring how long it took to issue those commands into buffer. So the Helix sees that scene is redrawing fine at high framerate but the graphics card is crying from all the work it is getting.

Anyway, if this one is the same as I experienced, it is really critical to solve as any complex scene is impossible to view.

holance:
I will look into this issue as well. While using mouse rotate, the OnRendering routine keeps falling into these two sections:
https://github.com/helix-toolkit/helix-toolkit/blob/master/Source/HelixToolkit.Wpf.SharpDX/Controls/DPFCanvas.cs#L614
https://github.com/helix-toolkit/helix-toolkit/blob/master/Source/HelixToolkit.Wpf.SharpDX/Controls/DPFCanvas.cs#L628

It may be some kind of timing issue and queued up the dispatcher or the composite renderer.

holance:
Close issue for now. This particular sluggish issue for me has been fixed in #331. @BetaRavener please open anther issue if you still experience the performance issue.

PR #331, opened 2016-11-07
Potential fix for sluggish during mouse rotation every few seconds(#327)

holance:
#327

Potential fix for sluggish during mouse rotation every few seconds.

I think the underlying issue for the sluggish is because the rendering is done by Rendering thread(CompositeTarget), but in some cases, the code also use Dispatcher to do rendering by using UI thread. Seems like there is a performance impact if keep switching between these two renderings. I just removed the dispatcher rendering part and only uses rendering thread. 

Need help to do some test or a code review. Please let me know the results.


BetaRavener:
I can try this patch on my scene that was being sluggish later today or tomorrow. The code looks ok at first glance, although I'm not sure it there is problem only with mouse camera rotation. If there are more code paths affected by this, maybe the discarding of event shouldn't be done like it is now in `MouseGestureHandler.cs` but in renderer. However, it should be sufficient for testing if the problem was resolved by this.

holance:
Make sure to include my latest checkin, seems like it fixes the FPS overshoot(now keeps at 120 fps max) during mouse rotation.

BetaRavener:
I built the latest version of your fork. Doesn't look like it solves my problem. I'll try to make screen capture tomorrow to illustrate what I'm experiencing.

holance:
It solved my problem very well. What's the number for
'lastRenderingDuration' on your side? There are still some performance
issue if it runs on my 3 years old laptop. But runs well on my desktop. I
think there are still some spaces to tweak to improve the performance.

holance:
@BetaRavener These are some things I noticed during the performance investigation.
1. When I run the example on my 3 year old laptop, debug mode is actually more smooth than release mode during mouse rotation. I think the root cause is mouse move event fires too frequent, it may cause the rendering thread to busy to keep up. I limit the mouse event frequency in latest commit. Please try it and see if you have any performance gain for your test case.
2. MSAA is always set to the maximum the graphic card can support, but not the graphic card can handle for actual rendering. So I put the dependency property so user can manually set MSAA level. You can try to disable MSAA completely. 

Please let me know if any of these helps your issue.

BetaRavener:
Sorry I'm bit bussy these days, will try out in the next 2 days :) Thanks for all the work.

PR #339, opened 2016-11-11
Enhancement of Triangulator with holes

Iluvatar82:
The previously created faster Triangulator was now enhanced to also work with multiple Holes.
The Holes need to be positioned completely inside the Polygon-Boundary.
Multiple Holes may not overlap, otherwise the triangulation may not work!

It is now a little slower than before but the added functionality is worth it.


Iluvatar82:
A screenshot from the current triangulator:
triangulation_with_holes
The Triangulation of a circle with a randomized radius of each of the 1000 points with the two holes of also 1000 points each needed 1.5 seconds, which is still a lot faster than the naive triangulator from before (which didn't support holes at all)

PR #335, opened 2016-11-08
TubeVisual3D, TorusVisual3D and MeshBuilder enhanced

Iluvatar82:
The TubeVisual3D has additional DP AddCaps to be able to automatically add caps to the ends of the tubes (#326)

MeshBuilder:
The AddTorus was enhanced and made a closed triangle-mesh even if self-intersecting (#332)
AddPyramid added optional Argument “closeBase” to be able to also generate two triangles at the base of the triangle
AddTetrahedron, AddOctahedron, AddDodecahedron implemented to be able to generate the remaining (not yet implemented) platonic solids (#35)

The MeshBuilder enhancements were also implemented in SharpDX

Issue #332, opened 2016-11-07
TorusVisual3D enhancement (close the mesh)


Iluvatar82:
I think maybe it would be nice to to remove the duplicate points and triangles in the TorusVisual3D. What do you think?

Iluvatar82:
Solved in pull request #355

PR #341, opened 2016-11-13
Ability to add caps in HelixToolkit.SharpDx meshbuilder. Fix caps normal issue in TubeVisual in WPF version

holance:
Added ability to add caps in HelixToolkit.SharpDx meshbuilder. 

Fix caps normal issue in TubeVisual in WPF version

### DIFF
--- a/Source/HelixToolkit.Wpf.SharpDX/Utilities/MeshBuilder.cs
+++ b/Source/HelixToolkit.Wpf.SharpDX/Utilities/MeshBuilder.cs
@@ -2836,15 +2836,16 @@ namespace HelixToolkit.Wpf.SharpDX
             }
 
             this.AddRectangularMeshTriangleIndices(index0, pathLength, sectionLength, isSectionClosed, isTubeClosed);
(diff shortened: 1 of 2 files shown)


Iluvatar82:
As mentioned in issue #332, this was already solved in pull request #335! I solved it not in the mesh builder but in the file "HelixToolkit.Wpf/Visual3Ds/MeshVisuals/TubeVisual3D.cs" because it's a property of this class only. This pr is redundant.

holance:
I know, but your solution did not specify the cap vertex normal properly,
causes lightening issue. I also added the cap in sharpdx version.

Iluvatar82:
Ok then, I just thought a solution in Wpf was needed, and since no equivalent to the TubeVisual3D exists in SharpDX I decided to implement it in the Tube's Tesselate function (as you mentioned in the issue #332 I think).

holance:
Yes, maybe my check in notes are a little bit confusing.  I added the cap function in sharpdx meshbuilder. And corrected the vertex normal issue in WPF TubeVisual tesselation. :)

PR #349, opened 2016-11-14
Move triangulation code to shared project

objorke:


objorke:
@Iluvatar82 I think it will be easier to maintain the triangulation code if it is moved to a shared project

Iluvatar82:
Ok, @objorke, you handle the transition :), I will have a look into the changes tomorrow (I hope).
Thank you for helping with the shared projects!

PR #358, opened 2016-11-18
Added FixedRotationPoint and FixedRotationPointEnabled properties for Viewport3DX. Billboard background color for BillboardSingleText3D

holance:
Added FixedRotationPoint and FixedRotationPointEnabled properties for Viewport3DX under CameraMode=Inspect. User can manually define a fix rotation point irrelevant to current camera lookat and location.


holance:
Add background color support for BillboardSingleText3D

PR #379, opened 2016-11-30
Added ReuseVertexArrayBuffer property to reuse existing DefaultVertex array

holance:
1. Added ReuseVertexArrayBuffer property to reuse existing DefaultVertex array and reduce memory allocation during realtime geometry/texture changes.
2. Remove unnecessary new struct object during struct array element operation.


holance:
Seems like build server gets stuck. Let me create a new pull request and try again.

PR #381, opened 2016-11-30
Reverted a change that is causing stuttering in camera rotation animations

oyvindse:
In addidtion to animation stuttering I also experienced FPS drops in complex scenes. It is hardly recognizable in the Helix examples, but there are some stuttering during the stop animation when releasing the mouse button while rotating. I went from 40 to 60fps in one of my own projects after this. Looks like it was first implemented in commit d236ded08d6dcb7b13597e11f622687bc1e1078e.

### DIFF
--- a/Source/HelixToolkit.Wpf.SharpDX/Controls/DPFCanvas.cs
+++ b/Source/HelixToolkit.Wpf.SharpDX/Controls/DPFCanvas.cs
@@ -648,23 +648,18 @@ namespace HelixToolkit.Wpf.SharpDX
             {                
                 if (pendingValidationCycles > 0)
                 {
-                    var cycle = System.Threading.Interlocked.Decrement(ref pendingValidationCycles);
+                    System.Threading.Interlocked.Decrement(ref pendingValidationCycles);
 
                     var t0 = renderTimer.Elapsed;
 
                     if (surfaceD3D != null && renderRenderable != null)
                     {
-                    // Update all renderables before rendering 
-                    // giving them the chance to invalidate the current render.                                                            
+                        // Update all renderables before rendering 
+                        // giving them the chance to invalidate the current render.                                                            
                         renderRenderable.Update(t0);
-                        if (cycle == 1)
-                        {
-                            Render();
-                        }
-                        else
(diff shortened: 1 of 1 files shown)


holance:
Revert this will cause unnecessary rendering twice, need more investigation. Would you provide more details on when you experience the shuttering? Thanks.

holance:
@oyvindse Please try to replace the InvalidateRedner() in DPFCanvas along with your change and see if there is any issue. Not sure why it is set to do 2 cycle rendering before my changes( @objorke Would you comment on this? Thanks.). If one cycle is working fine, we can change to one cycle.


        private void InvalidateRender()
        {
            //Use pendingInvalidateOperation to check if there is a pending operation.
            if (pendingInvalidateOperation != null)
            {
                switch (pendingInvalidateOperation.Status)
                {
                    case DispatcherOperationStatus.Pending:
                        System.Threading.Interlocked.CompareExchange(ref pendingValidationCycles, 1, 0);
                        return;
                    case DispatcherOperationStatus.Executing:
                        pendingInvalidateOperation.Abort();
                        break;
                }
                pendingInvalidateOperation = null;
            }
            // For some reason, we need two render cycles to recover from 
            // UAC popup or sleep when MSAA is enabled.
            if (System.Threading.Interlocked.CompareExchange(ref pendingValidationCycles, 1, 0) != 0)
            {
                //If invalidate failed, schedule an async operation to invalidate in future
                pendingInvalidateOperation
                    = Dispatcher.BeginInvoke(invalidAction, DispatcherPriority.Background);
            }
        }

oyvindse:
Thanks for reviewing @holance. I Tried to replace InvalidateRender() as you suggested, and I can't find any issues when running with one cycle.

holance:
Hi oyvindse, I merged the code with RenderCycle set to 1 at #382. Hope it solves your issue.

oyvindse:
Great @holance . This does indeed solve my issue. Btw, there's a small typo in the registering of the dependency property. You have accidently called it "RenderCylces".

holance:
Close this pull, since it has been addressed in #382

Issue #407, opened 2017-01-08
DataTemplate example doesn't work properly


Fendoroid:
There's a major problem with this example: bound properties are set correctly, but non bound properties are not set at all.
To better analyze the problem, i did some edits. I commented out the `FixedElements`'s template:
```
<!--<local:ItemsVisual3D ItemTemplate="{StaticResource Template1}" ItemsSource="{Binding FixedElements}"/>-->
```

And I edited the data template for the `CubeVisual3D`:
```
<local:DataTemplate3D x:Key="Template2">
            <h:CubeVisual3D Center="{Binding Position}" SideLength="2" Fill="OrangeRed"/
</local:DataTemplate3D>
```
With this configuration I still got blue cubes of size 1. Plus, they are strangely misaligned.

However, uncommenting the commented block of code in `DataTemplate3D.cs` apparently sets every property correctly on debugger if you break right before `CreateItem` returns. Strangely, the output is not correct (`Fill` and `SizeLength` are set properly, but not `Center`, that appears to be at the origin).

In all cases, appears the following error:
```
System.Windows.Data Error: 2 : Cannot find governing FrameworkElement or FrameworkContentElement for target element. BindingExpression:Path=Position; DataItem=null; target element is 'CubeVisual3D' (HashCode=61946665); target property is 'Center' (type 'Point3D')
```

I don't know what's happening here, sometimes the results are pretty unpredictable. I think those problems arises because of the binding error. However, I've no idea of how to resolve it.

objorke:
Right, only properties with very simple bindings are currently supported in this example:
https://github.com/helix-toolkit/helix-toolkit/blob/develop/Source/Examples/WPF/ExampleBrowser/Examples/DataTemplate/DataTemplate3D.cs#L46
It should be easy to add support for 'static' values, but still there will be challenges with other types of bindings...

Fendoroid:
The binding error is caused by the binding defined in xaml, the `Content` property of the data template. I still don't know why it shows up, if `CubeModel3d` derives from `Visual3D` which is a `DependencyObject`, that is binding-ready.
If you do a `BindingOperation.ClearBinding` to each bounded property in the `Content` of `DataTemplate3D` the error disappears, but there is no advantage in doing that, since it removes the further binding capabilities of the template.

I still don't get why binding gets invalidated if I set non-bounded properties. That happens only with the foreach loop, not if I set manually single properties like:

```
[...]

PR #420, opened 2017-01-14
Support complex template objects as DataTemplate3D with bindings on any level.

ccHanibal:
I rewrote the whole DataTemplate3D logic a while ago, but didn't commit it somewhere. Now there is a ticket #407 and I remebered having that done before. So here it is.

The cloning now takes place at XAML level (serialising and deserialising), so all bindings and direct assignments will be set correctly. The bindings are then updated with the data context object as its source in the whole tree.

I also included a DataTemplateSelector3D to support dynamic templating based on the type of the data context. For this to work the DataTemplate3D's key must be set to the type of the target object.

The only downside is the performance. Inserting 250 objects at once takes abount 1 second.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -15,6 +15,7 @@ ando23 <andreas@herzig-net.de>
 Benglin (Ben) Goh <benglin@outlook.com>
 Brad Phelan <bradphelan@xtargets.com>
 Carl Philipp Bickmeier <c.bickmeier@diomex.de>
+ccHanibal <oliver91daum@hotmail.de>
 christophano@hotmail.com
 Christof Konstantinopoulos <chrkon@web.de>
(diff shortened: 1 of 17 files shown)


KKauK:
Hi ccHanibal, this works pretty well, but I stumbled into a new problem.
When using binding together with a converter the binding is not recognized and so will fail.

Position="{Binding CenterPoint, Converter={StaticResource PointConverter}}"

Appreciate your help!

ccHanibal:
@KKauK can you please open an issue for that? This PR was merged long ago and you question is not about the PR itself.
But in short I think this is pretty difficult. A static resource lives in a ResourceDictionary. We would have to merge all existing in the xaml of the Model3D. On the otherhand can't you use an inline converter via MarkupExtension class?

PR #438, opened 2017-02-01
Move octree feature from shared project to Wpf.SharpDX

JeremyAnsel:
Since the octree feature doesn't compile in UWP project, move it to Wpf.SharpDX.


JeremyAnsel:
Added the UWP project to HelixToolkit.AppVeyor solution.

PR #440, opened 2017-02-07
Improve event skipper to increase FPS rate

holance:
Improve event skipper to increase FPS rate
Added MaxFPS dependency property to allow user set desired FPS


holance:
The FPS does not reflect the actual FPS rate for 3D rendering. Need to move the FPS counter into DPFCanvas, will fix later.

PR #466, opened 2017-03-31
Bug fixes and improvements

holance:
1. Fix octree FindNearestPointBySphereExcludeChild triangleIndices in hitResult.
2. Move create bounding box/bounding sphere into Geometry3D.
3. Avoid calling update visuals when text changed. Only needs to rebuild visuals on ModelUpDirection changed.
4. Unproject bug fix, cannot use render context view matrix. Causes panning issue

### DIFF
--- a/Source/HelixToolkit.Wpf/Visual3Ds/Composite/ViewCubeVisual3D.cs
+++ b/Source/HelixToolkit.Wpf/Visual3Ds/Composite/ViewCubeVisual3D.cs
@@ -28,13 +28,19 @@ namespace HelixToolkit.Wpf
         /// Identifies the <see cref="BackText"/> dependency property.
         /// </summary>
         public static readonly DependencyProperty BackTextProperty = DependencyProperty.Register(
-            "BackText", typeof(string), typeof(ViewCubeVisual3D), new UIPropertyMetadata("B", VisualModelChanged));
+            "BackText", typeof(string), typeof(ViewCubeVisual3D), new UIPropertyMetadata("B", (d,e)=> 
+            {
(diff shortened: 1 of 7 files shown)


Source/HelixToolkit.Wpf/Visual3Ds/Composite/ViewCubeVisual3D.cs
```diff
@@ -30,6 +30,7 @@ public class ViewCubeVisual3D : ModelVisual3D
         public static readonly DependencyProperty BackTextProperty = DependencyProperty.Register(
             "BackText", typeof(string), typeof(ViewCubeVisual3D), new UIPropertyMetadata("B", (d,e)=> 
             {
+                var b = (d as ViewCubeVisual3D).GetCubefaceColor(1);
                 (d as ViewCubeVisual3D).UpdateCubefaceMaterial(1, Brushes.Red, e.NewValue == null ? "" : (string)e.NewValue);
```
JeremyAnsel: Should it be `(d as ViewCubeVisual3D).UpdateCubefaceMaterial(1, b, e.NewValue == null ? "" : (string)e.NewValue);`?

JeremyAnsel: Should it be `(d as ViewCubeVisual3D).UpdateCubefaceMaterial(1, b, e.NewValue == null ? "" : (string)e.NewValue);`?

holance:
Yes, i will fix it. Thanks:)

Source/HelixToolkit.Wpf/Visual3Ds/Composite/ViewCubeVisual3D.cs
holance: Fixed

PR #481, opened 2017-05-16
Added LineHitTestResult class with line-specific information about the hit

smischke:
The scalar of the closest point on line ("Tc") is useful if you need to know at which percentage of the line the hit was.

### DIFF
--- a/Source/HelixToolkit.Wpf.SharpDX/Controls/HitTestResult.cs
+++ b/Source/HelixToolkit.Wpf.SharpDX/Controls/HitTestResult.cs
@@ -66,6 +66,11 @@ namespace HelixToolkit.Wpf.SharpDX
     public class LineHitTestResult : HitTestResult
     {
         /// <summary>
+        /// Gets or sets the index of the line segment that was hit.
+        /// </summary>
+        public int LineIndex { get; set; }
+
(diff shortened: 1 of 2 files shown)


holance:
One concern about changing the distance to LineToRay distance for LineGeometryModel3D is that for all other models, the distance represents the Ray origin to hit point distance. 

When the hit test returns, it is sorted by the distance. Then the result will be wrong order. You can test the hit test in Instances demo. It shows the hit test combination between MeshGeometryModel3D and LineGeometryModel3D hit tests.

Maybe you can add another LineToRay distance property in LineHitTestResult to separate these two distances? Thanks.

smischke:
You're absolutely right, thanks for reviewing. I fixed it.

smischke:
Maybe it would be even better to put `TriangleIndices` into a specialized `MeshHitTestResult`.

smischke:
Or changing 
`public System.Tuple<int, int, int> TriangleIndices { set; get; }` 
to
`public IList<int> HitGeometryIndices { get; set; }` in `HitTestResult`?

holance:
Looks good. Thanks for your contribution. We can change the TriangleIndices later if necessary.

PR #487, opened 2017-06-07
Fix XAML incompatibility while keeping most of the "anti DP overhead stuff"

smischke:
Your recent changes to reduce Dependency Property overhead broke XAML compatibility. Please consider applying this pull request to restore it. ExampleBrowser-10044 was reverted to it's original state to demonstrates the issue.


smischke:
There are other problems, too with the recent changes. I have to investigate this more thoroughly.

smischke:
The "other problems" were in my application. Please consider applying this pull request to restore XAML compatibility.

holance:
For GroupElement3D, Please see my comment on @485

Removed [ContentProperty("Children")] attribute from GroupElement3D.
( Since the children in GroupElement3D is used for binding, if use [ContentProperty("Children")] and add children in xaml, causes all children goes into the same static collection.
To properly use [ContentProperty("Children")], children needs to use read-only DependencyPropertyKey or normal property like in CompositeModel3D.)

holance:
GroupElement3D has been properly implemented in #491 . Fixed the demo bugs as well.  Please use ItemsSource in GroupElement3D for collection binding.

Thanks for your contribution.

PR #498, opened 2017-06-25
Fixed null reference exception if point geometry is empty

Fellfalla:
Case: `geometryInternal ` is null.

Reproduction: 
- Create empty instance of PointGeometry3D 
- load model into the Viewport
- click into the view.

Result: 
-> Crash


Fellfalla:
Meanwhile, I've done it. Should i send you the pull request?

PR #502, opened 2017-06-28
fixed crash bug due to undefined model bounds

Fellfalla:
Reproduction:
- Create a ViewportScene with a Empty MeshGeometryVisual3D (The model.Bounds have to be undefined)
- Toggle between the Viewport and another tab forth and back

Result:
- Program termination

### DIFF
--- a/Source/HelixToolkit.Wpf.SharpDX/Controls/ViewportExtensions.cs
+++ b/Source/HelixToolkit.Wpf.SharpDX/Controls/ViewportExtensions.cs
@@ -192,7 +192,7 @@ namespace HelixToolkit.Wpf.SharpDX
             foreach (var element in viewport.Renderables)
             {
                 var model = element as IBoundable;
-                if (model != null)
+                if (model != null && !model.Bounds.Minimum.IsUndefined() && !model.Bounds.Maximum.IsUndefined())
                 {
                     if (model.Visibility != Visibility.Collapsed)
                     {


holance:
Hi Fellfalla, thanks for your contribution.
I am not able to reproduce the problem you described, would you provide an example so we can test? Thanks.

Fellfalla:
try to create a ViewportDX and set the property `ZoomExtentsWhenLoaded="True"`.

Put this in one tab of your main view:
```
<hx:Viewport3DX IsPanEnabled="True"
                ZoomExtentsWhenLoaded="True"
                IsInertiaEnabled="True"
                IsManipulationEnabled="True"
                IsTouchZoomEnabled="True"
                ViewCubeVerticalPosition="Top"
                ViewCubeOpacity="0.5">
    <hx:MeshGeometryModel3D Geometry="{Binding GeometryProperty}"/>
</hx:Viewport3DX>
```

and add in the ViewModel

```
public MeshGeometry3D GeometryProperty
{
    get
    {
        var builder = new MeshBuilder();
        return builder.ToMeshGeometry3D();
    }
}
```

or maybe

```
public MeshGeometry3D GeometryProperty
{
    get
    {
        return null;
    }
}
```

Then run and switch between the tab with the Viewport and another. I hope this reproduces it. If not, i'll setup a example.

holance:
I found the bug when the RenderHost is null. It has been fixed in #504, please let me know if this is fixed the issue you experienced.

Fellfalla:
No, unfortunately this does not help.

The undefined bound sizes do cause the error in the FindBounds method.
image

holance:
Would you be able to share the project? I still cannot figure out how this happened. The default BoundingBox values are supposed be all zeros instead of NaN.

Fellfalla:
Here we go: Now I know how i exactly produced the error.
I bound to a mesh that has been created somehow in this way:


    var meshBuilder = new MeshBuilder();

    meshBuilder.AddBox(
                 new Vector3(1, 0, 0),
                new Vector3(0, 0, 1),
                new Vector3(0, 1, 0),
                double.NaN, 1, 1);

    // Create the Geometry
    var mesh =  meshBuilder.ToMeshGeometry3D();

[HelixCrash.zip](https://github.com/helix-toolkit/helix-toolkit/files/1117802/HelixCrash.zip)

holance:
The double.nan is not valid. What's the reason you want to pass the NAN into the builder?

Fellfalla:
There is the chance to bind a `Geometry` to a 3D-Object, which dimensions or position is defined only after a certain time in the program. If this object is instanciated at the beginning and does contain `struct`s for its dimensions, then you have some `double.NaN`s as input parameter for a geometry building converter.

holance:
If allow to contain invalid position, we have to check if every position is valid, why not just bind to a null model and change the model at runtime?

Fellfalla:
You are right that binding a null modell is better practise. At least, for the programmer it is not clear why the program crashes, if he did insert double.NaN (especially if double.NaN is a mistake). In my case, the debugging would have taken a huge amount of time, if i had no acces to the source code. Especially the time the error arrices is not quite intuitive.

holance:
Close the pull request for now. Needs to improve exception throwing in future.

PR #612, opened 2018-02-05
Adds SortingGroupModel3D to HelixTollkit.Wpf.SharpDX, a models contai…

N4rk0:
…ner that sorts transparent items by their distance from camera in order to properly render them.
Implementation is almost the same of SortingVisual3D


holance:
Current version in Develop does not support VS2017 syntax yet. 
I am working on newer version of the toolkit under Feature2.0 branch (supports VS2017). If you are interested, you can start implement the model under new architecture. You can do the sorting in Update function.

N4rk0:
Ok, i'll try to add it to develop branch.

PR #641, opened 2018-02-24
Unstage cso files

JeremyAnsel:
Unstage shader compiler output cso files.


JeremyAnsel:
Make the SharpDX project depend on the ShaderBuilder project.
Add the ShaderBuilder project to the SharpDX and AppVeyor solutions.

PR #686, opened 2018-04-03
Fixed SweepLinePolygonTriangulator for a base case triangle.

ChaosSorcerer:
Added a demo and test for it.

I have to locally remove "using HelixToolkit.Wpf;" from Polygon3DTests, to get the tests to build. However as it's not my test class and I have not altered it, I am hesitant on changing that.


holance:
As long as it pass the build, it should be fine. Please fix the build error so it can be merged.

ChaosSorcerer:
Fixed the build, it an be merged now.

PR #687, opened 2018-04-04
SortingGroupModel3D - sorts items dpending on distance from the camera

N4rk0:
Allows correct rendering of transparent items


holance:
I will implement an updated one for the sorting group model in these two days.

holance:
Implemented in #691.

Issue #702, opened 2018-04-09
How to attach EnvironmentMap3D from code?


vvs551:
I'm testing the development version of HelixToolkit.SharpDX
And now I could not attach new EnvironmentMap3D from code:

 Viewport3DX modelView;
 EnvironmentMap3D EnviMap;
 ..           
 modelView.Items.Add(EnviMap);
 EnviMap.Attach(modelView.RenderHost);
 
ERROR: EnvironmentMap3D  does not contain definition for Attach!

How to add new EnvironmentMap3D  from code to Viewport3DX in this version of HelixToolkit.SharpDX?

holance:
You should not worrying about attach, it is handled automatically.

vvs551:
Is Items.Add(..) enough?

holance:
Just make sure you assign the effects manager after adding models into the viewport3dx.Items. Or add a GroupModel3D into viewport first, then adds model into the groupmodel3d.children at anytime.

holance:
I put an fix #703 so viewport3dx itself can add/remove model at run time. Please try again using Viewport3DX.Items.Add for your environment map. It should work.

holance:
Did you get the latest build? I put a tester(Viewport3DXCodeBehindTester) in #705. It works ok. If it is still not working, attach an example on how you initialize your viewport and I will take a look.

vvs551:
I fond the reason of empty screen!
If I use dds with size 60M and more - I see a black scree
Unfortunately, all my dds are more than this limit

holance:
Could you provide one of the large dds so I can try?

holance:
Seems like I can render your cubemap. But your cubemap is having some issue, the coordinate is wrong and not render correctly. But at least not black screen.

vvs551:
I used it in previous version of helixtoolkit
There is a screen shot
But now - black viewport (
![screenshot](https://user-images.githubusercontent.com/14919853/38669490-36a6c96e-3e4f-11e8-8b0d-94caf566c3bb.jpg)

vvs551:
Could you help to select a correct output format for .dds?
![image](https://user-images.githubusercontent.com/14919853/38674090-ef51a900-3e5b-11e8-8660-d5ca9d2240c6.png)

holance:
Please try again. If still not render correctly, reopen this issue.

vvs551:
Problem was out

PR #703, opened 2018-04-09
 Let viewport3dx supports add/remove model at run time.

holance:
Let viewport3dx supports add/remove model at run time. #702
Always return new material object from material factory.


vvs551:
Could you get any code sample: how to add model to viewport3dx runtime?

PR #820, opened 2018-06-28
Ply model importer

Bez99:
I've added some code in the "Importers" directory for reading some elements of a .ply file.


holance:
I tried to load some models from http://graphics.stanford.edu/data/3Dscanrep/, but some are not getting loaded and throws exception(such as [XYZ RGB dragon.ply.gz](http://graphics.stanford.edu/data/3Dscanrep/xyzrgb/xyzrgb_dragon.ply.gz)). Could you help to fix some loading issues? Thanks.

PR #842, opened 2018-07-11
Visuals Point/Rect selection

### LINKED ISSUE #841: Rectangle selection: why returns Models3D instead of Visuals3D?
I don't understand why the SelectionCommands are returning ModelsSelected (list of Model3D) instead of a list of Visual3D.
IMHO the "selection" operation is focused on Visuals not on Models.
Citing "ModelVisual3D vs. Model3DGroup — When to use which?" https://blogs.msdn.microsoft.com/danlehen/2005/10/09/modelvisual3d-vs-model3dgroup-when-to-use-which/

"You can hit test a Visual."
"Visuals are single-use.  Content can be shared/reused between Visuals."
"Visuals support services like hit testing.  Content are serialized rendering instructions and nothing else."

In my project using HelixToolkit i've done a little "quick and dirty" experiment to reproduce all the Viewport3DHelper.FindHits stack modifying all Traverse funcions to use an Action<T, Visual3D, Transform3D> callback to manage the selection of the Visuals with the models and it seems to work.

Now i'm guessing if i'm on the wrong way, or if this modification may be appreciated by others and if wrap-up my work and make a PR.

beppemarazzi:
Ability to select not only Models but also Visuals with RectangleSelectionCommand and PointSelectionCommand.
In XXXSelectionCommand ctor you can specify an handler for SelectionModelsEvent (as before) or an handler for SelectionVisualsEvent or both.
Extended all Traverse<> funcions to pass the visual all around.
Updated the example "RectSelection".
If ok can close #841

### DIFF
--- a/Source/Examples/WPF/ExampleBrowser/Examples/RectSelection/MainWindow.xaml
+++ b/Source/Examples/WPF/ExampleBrowser/Examples/RectSelection/MainWindow.xaml
@@ -6,6 +6,8 @@
     <DockPanel>
         <StackPanel DockPanel.Dock="Top" Orientation="Horizontal" Margin="7, 5">
             <ComboBox SelectedValue="{Binding SelectionMode}" ItemsSource="{Binding SelectionModes}"/>
+            <TextBlock Text="SelectedVisuals:"/>
+            <TextBlock Text="{Binding SelectedVisuals}"/>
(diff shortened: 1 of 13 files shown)


holance:
Seems like there is an comment error during the build

PR #890, opened 2018-08-07
Add Axis aligned Plane Grid Model

holance:
Add Axis aligned Plane Grid Model. Similar to the plane pattern in 3D Builder in windows. Supports tile or grid patterns.


holance:
image
image

PR #905, opened 2018-08-17
Restore TypeConverter attributes in DependencyObjects to make it possible again to define i.e. material colors in XAML (WPF).

smischke:


holance:
Maybe I will just change all color in material to windows.media.color?

holance:
Never mind. Too many things need to be changed.

Issue #1028, opened 2018-12-11
Improve rotation around mouse down point


ChaosSorcerer:
When setting the helix viewport settings RotateAroundMouseDownPoint="true", ZoomAroundMouseDownPoint="true" and  CameraMode="Inspect", the camera correctly rotated around and zooms towards a point on a 3D object under the mouse cursor. However when there is no object below the mouse cursor it project the 2D location where the mouse hits to a 3D location on the same plane as the camera's current target location. 

This seems fine, untill the user rotates around and zoom back and forth objects a lot. The camera's target position can be placed arbitrarily close or very far away from the camera. Thus whenever an object is not selected we get problemetic user expierences such as:
- zooming doing next to nothing 
- instantly zooming past the entire scene
-  The camera rotating around a point directly in front of it, thus rotating away from the scene rather than around it. 

As a result I'd like to change this functionality to rotating around the closest 3D point in the scene. I've made a branch, made some changes and tested things out and it is now functioning in a way that is easier to understand than the old method (I can add a toggle to swap between the old an the new version if needed). Have not tested performance on scenes with huge ammounts of vertices yet though.

PR #1029, opened 2018-12-11
When RotateAroundMouseDownPoint and ZoomAroundMouseDownPoint are both…

ChaosSorcerer:
When RotateAroundMouseDownPoint and ZoomAroundMouseDownPoint are both on the camera target position may get really close/very far from the camera, causing bad/unexpected camera interaction when not hovering over an object. See https://github.com/helix-toolkit/helix-toolkit/issues/1028 for more information.

Added a Closest3DPointHitTester and used it to allow rotating around the closest vertex instead of always using the unprojected position when no meshes are under the mouse curser.

### DIFF
--- a/Source/Examples/WPF/SimpleDemo/MainWindow.xaml
+++ b/Source/Examples/WPF/SimpleDemo/MainWindow.xaml
@@ -229,7 +229,7 @@
             </Style>
         </Grid.Resources>
         <!--  The HelixViewport3D supports camera manipulation, and can be used just like the Viewport3D  -->
-        <HelixToolkit:HelixViewport3D ShowFrameRate="True" ZoomExtentsWhenLoaded="True">
+        <HelixToolkit:HelixViewport3D ShowFrameRate="True" ZoomExtentsWhenLoaded="True" ZoomAroundMouseDownPoint="True" RotateAroundMouseDownPoint="True">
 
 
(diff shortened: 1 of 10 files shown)


holance:
If finding closest vertex is too costly, maybe just find the closest Model3D Bound or the center of the Bound?

ChaosSorcerer:
Added the option to use the closest model3D bound on large meshes.

ChaosSorcerer:
Also made a small addition to avoid zooming and rotating around vertices that are not currently visible within the view

hansmei:
How about adding this code for the WPF.SharpDX CameraController as well?

holance:
Usually I would suggest to use FixedRotationPoint and do your own HitTest and find the Point you want to rotate. Using RotateAroundMouseDownPoint/ZoomAroundMouseDownPoint are not really a reliable way to control this kind of behavior since you may use mouse down on viewport for many other different behaviors.

hansmei:
Many applications are using the zoom event to set a pivot point (preferably with a ball marker) on hit for which the rotation would rotate around.

That is the kind of behaviour I'm missing so far with the current abilities of the WPF.SharpDX project. I have tried to tweak many different settings, but setting a pivot (fixed rotation point) does not seem to work the way I want to. An ability to set a pivot point on zoom for the rotation would be great.

PR #1066, opened 2019-01-14
 add AddCircle method for LineBuilder

NaBian:


holance:
Function has been merged in #1069, with position/normal as parameters.

PR #1217, opened 2019-08-23
Fixed a bug in the Ply Reader.

B3zaleel:
The ply reader now handles big endian ply files better, though loading time might be a bit slow. The mode of parsing ply files has also been updated to allow support of non-mainstream elements and properties.


Source/HelixToolkit.Wpf.Shared/Importers/PlyReader.cs
```diff
@@ -13,12 +13,8 @@ namespace HelixToolkit.Wpf
     /// <summary>
     /// Polygon File Format Reader.
     /// </summary>
-    /// https://www.cc.gatech.edu/projects/large_models/ply.html
-    /// http://graphics.stanford.edu/data/3Dscanrep/
-    /// <remarks>
-    /// This reader only reads ascii ply formats.
-    /// This was initially meant to read models exported by Blender 3D Software.
-    /// </remarks>
+    /// <see cref="https://www.cc.gatech.edu/projects/large_models/ply.html"/>
```
holance: Seems like this is giving error.

Source/HelixToolkit.Wpf.Shared/Importers/PlyReader.cs
```diff
@@ -13,12 +13,8 @@ namespace HelixToolkit.Wpf
...
+    /// <see cref="http://graphics.stanford.edu/data/3Dscanrep/"/>
```
holance: This is having error too. Maybe remove the <see cref/>

PR #1275, opened 2019-12-11
Fix appveyor build

JeremyAnsel:
When building with holance account, the build fails because the MyGet key belongs to the objorke account.

This PR deploys the packages to MyGet on objorke account only.
This fixes the build for holance account.


JeremyAnsel:
The failed build doesn't seem to be related to this PR.

holance:
Yes, seems like something is changed in build system recently.

PR #1318, opened 2020-02-25
HitTestThickness is not taken in account into the bounds computation …

mdelanno:
In the FindHits process, there is a step where the ray is tested against the bounds (`PreHitTestOnBounds()`). In the case of a vertical or an horizontal line, the bounds can be very thin (in fact the minimum size is fixed to 0.2) and the HitTestThickness is not taken in account when the bounds are computed so the test fails before the actual Ray/Line distance is computed.

I think we should skip the `PreHitTestOnBounds()` in the case of a line (by make it returns true) because the HitTestThickness is not taken in account .

### DIFF
--- a/Source/HelixToolkit.SharpDX.Shared/Model/Scene/LineNode.cs
+++ b/Source/HelixToolkit.SharpDX.Shared/Model/Scene/LineNode.cs
@@ -126,6 +126,12 @@ namespace HelixToolkit.UWP
             {
                 return (Geometry as LineGeometry3D).HitTest(context, totalModelMatrix, ref ray, ref hits, this.WrapperSource, (float)HitTestThickness);
             }
+
+            protected override bool PreHitTestOnBounds(ref Ray ray)
+            {
+                // HitTestThickness is not taken in account into the bounds computation so we can not use it for hit test.
+                return true;
+            }
         }
     }
 


holance:
I think we need different way to do this checking. It may affect hit performance by simply returning true.

holance:
I think we need different way to do this checking. It may affect hit performance by simply returning true.

mdelanno:
The only way I see is to use a 2d bounding box like this but I'm not sure it will be faster (I'm goind to test it):

    protected override bool PreHitTestOnBounds(ref Ray ray)
    {
        if (RenderHost == null || Geometry?.Positions == null) return true;

        float minX = float.MaxValue;
        float maxX = float.MinValue;
        float minY = float.MaxValue;
        float maxY = float.MinValue;
        foreach (Vector3 position in Geometry.Positions)
        {
            Vector2 vector2 = RenderHost.Viewport.Project(position);
            minX = Math.Min(minX, vector2.X);
            maxX = Math.Max(maxX, vector2.X);
            minY = Math.Min(minY, vector2.Y);
            maxY = Math.Max(maxY, vector2.Y);
        }

        minX -= (float)HitTestThickness;
        maxX += (float)HitTestThickness;
        minY -= (float)HitTestThickness;
        maxY += (float)HitTestThickness;

        Vector2 rayPosition = RenderHost.Viewport.Project(ray.Position);

        return minX <= rayPosition.X && rayPosition.X <= maxX 
            && minY <= rayPosition.Y && rayPosition.Y <= maxY;
    }

holance:
I will try to figure something out this weekend.

holance:
I think we can simply do a `BoundsSphereWithTransform.Intersects(ref ray)` only for line models.

mdelanno:
Sounds like a good compromise.

holance:
Relax the line pre-hit test with only bounding sphere. Ref #1319

PR #1324, opened 2020-03-13
Take cross section from planes into account when hit testing

eriove:
First of all, thanks for a great project!

**Some background on this pull request:**
We are showing 3D drawings with HelixToolkit and to see the "inside" of objects we use the `CrossSectionMeshGeometryModel3D` to cut at arbitrary locations along the X, Y and Z axes. We noticed that if a cutting plane is active the hit testing doesn't take that into account. This obviously makes it hard for the user to select objects behind other objects.

This pull request takes the cutting planes into account for hit testing for `CrossSectionMeshGeometryModel3D`. There are still **some open questions** from my side:

- [x] `CrossSectionMeshGeometryModel3D` doesn't exist in HelixToolkit.SharpDX.Core but rather in HelixToolkit.SharpDX.Core.Wpf. This means that the current solution doesn't work in .NET Core (see changes in `MeshGeometry3D`). Do you know how this could be resolved in a good way? I'm not sure if I understand the reasoning between the differences between the .NET Framework and .NET Core nuget package/project structure, but I guess it is related to the continuous changes from MS in what's available in .NET Core/Standard.
- [ ] I modified an existing example to test this, should I create a new one instead?
- [ ] What's the performance requirements, do I risk making the hit testing slower?
- [ ] Is there a use case for not taking the cutting planes into account? Will this break code for someone? Should it be optional?
- [ ] Tests for when OctTree is used?
- [ ] Test for WPF (without SharpDX)? Is there a way to use the cutting planes in WPF without SharpDX? I couldn't find it.

### DIFF
--- a/Source/Examples/WPF.SharpDX/CrossSectionDemo/MainWindow.xaml.cs
+++ b/Source/Examples/WPF.SharpDX/CrossSectionDemo/MainWindow.xaml.cs
@@ -31,5 +31,10 @@ namespace CrossSectionDemo
                 }
             };
         }
+
+        private void geometry_Mouse3DDown(object sender, HelixToolkit.Wpf.SharpDX.MouseDown3DEventArgs e)
+        {
+
+        }
     }
 }
--- a/Source/HelixToolkit.SharpDX.Core.Tests/HelixToolkit.SharpDX.Core.Tests.csproj
+++ b/Source/HelixToolkit.SharpDX.Core.Tests/HelixToolkit.SharpDX.Core.Tests.csproj
@@ -7,9 +7,8 @@
     </PropertyGroup>
 
     <ItemGroup>
-        <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.8.0" />
-        <PackageReference Include="MSTest.TestAdapter" Version="1.3.2" />
-        <PackageReference Include="MSTest.TestFramework" Version="1.3.2" />
+        <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.0.1" />
+        <PackageReference Include="NUnit" Version="3.12.0" />
     </ItemGroup>
 
     <ItemGroup>
--- a/Source/HelixToolkit.SharpDX.Shared/Model/Geometry/MeshGeometry3D.cs
+++ b/Source/HelixToolkit.SharpDX.Shared/Model/Geometry/MeshGeometry3D.cs
@@ -160,6 +160,12 @@ namespace HelixToolkit.UWP
             }
         }
 
+        /// <summary>
+        /// Callers should set this property to true before calling HitTest if the callers need multiple hits throughout the geometry.
+        /// This is useful when the geometry is cut by a plane.
+        /// </summary>
+        public bool ReturnMultipleHitsOnHitTest { get; set; }
+
         public virtual bool HitTest(RenderContext context, Matrix modelMatrix, ref Ray rayWS, ref List<HitTestResult> hits, object originalSource)
         {
             if(Positions == null || Positions.Count == 0
@@ -192,11 +198,7 @@ namespace HelixToolkit.UWP
                 {
                     int index = 0;
                     float minDistance = float.MaxValue;
-#if !NETFX_CORE
-                    bool isCrossSectionMeshGeometryModel3D = originalSource is CrossSectionMeshGeometryModel3D;
-#else
-                    bool isCrossSectionMeshGeometryModel3D = false;
-#endif
+
                     foreach (var t in Triangles)
                     {
                         var v0 = t.P0;
@@ -206,7 +208,7 @@ namespace HelixToolkit.UWP
                         {
                             // For CrossSectionMeshGeometryModel3D another hit than the closest may be the valid one, since the closest one might be removed by a crossing plane
                             if (isHit && result.IsValid &&
-                                isCrossSectionMeshGeometryModel3D)
+                                ReturnMultipleHitsOnHitTest)
                             {
                                 hits.Add(result);
                                 result = new HitTestResult
@@ -215,7 +217,7 @@ namespace HelixToolkit.UWP
                                 };
                             }
 
-                            if (d >= 0 && (d < minDistance || isCrossSectionMeshGeometryModel3D)) // If d is NaN, the condition is false.
+                            if (d >= 0 && (d < minDistance || ReturnMultipleHitsOnHitTest)) // If d is NaN, the condition is false.
                             {
                                 minDistance = d;
                                 result.IsValid = true;
--- a/Source/HelixToolkit.SharpDX.Shared/Model/Geometry/MeshGeometry3D.cs
+++ b/Source/HelixToolkit.SharpDX.Shared/Model/Geometry/MeshGeometry3D.cs
@@ -192,6 +192,11 @@ namespace HelixToolkit.UWP
                 {
                     int index = 0;
                     float minDistance = float.MaxValue;
+#if !NETFX_CORE
+                    bool isCrossSectionMeshGeometryModel3D = originalSource is CrossSectionMeshGeometryModel3D;
+#else
+                    bool isCrossSectionMeshGeometryModel3D = false;
+#endif
                     foreach (var t in Triangles)
                     {
                         var v0 = t.P0;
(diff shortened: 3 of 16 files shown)


Source/Examples/WPF.SharpDX/CrossSectionDemo/MainWindow.xaml.cs
```diff
@@ -31,5 +31,10 @@ public MainWindow()
                 }
             };
         }
+
```
eriove: This was a quick check to test the hit testing. I can move this to a new example or bind this to a command in the view model instead.

Source/HelixToolkit.Wpf.SharpDX.Tests/Elements3D/MeshGeometryModel3DTests.cs
```diff
@@ -9,20 +9,24 @@
...
 {
     [TestFixture]
+    [Apartment(ApartmentState.STA)]
     class MeshGeometryModel3DTests
     {
         /// <summary>
         /// Test for pull request #54.
         /// Fix IndexOutOfRange exception in CreateDefaultVertexArray.
         /// </summary>
-        [Test, STAThread]
```
eriove: The tests didn't run in a STA Thread with this attribute, changed it to `[Apartment(ApartmentState.STA)]`

holance:
HelixToolkit.SharpDX.Core should use `Scene Node` classes instead of Model3D. Model3D only available for WPF with dependency property. All Model3D class are wrapper for `Scene Node` classes.

Source/HelixToolkit.SharpDX.Shared/Model/Geometry/MeshGeometry3D.cs
```diff
@@ -192,14 +192,30 @@ public virtual bool HitTest(RenderContext context, Matrix modelMatrix, ref Ray r
                 {
                     int index = 0;
                     float minDistance = float.MaxValue;
+#if !NETFX_CORE
```
holance: I don't think this is gonna work.
SharpDX.Core.Wpf project depends on SharpDX.Core. 
You may need to add a separate paramter in HitTest function.

eriove: Sounds good, I was a bit reluctant to make too many changes, but that seems like the only reasonable way.

Source/HelixToolkit.SharpDX.Core.Tests/HelixToolkit.SharpDX.Core.Tests.csproj
```diff
@@ -7,9 +7,8 @@
     </PropertyGroup>
 
     <ItemGroup>
-        <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.8.0" />
-        <PackageReference Include="MSTest.TestAdapter" Version="1.3.2" />
-        <PackageReference Include="MSTest.TestFramework" Version="1.3.2" />
+        <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.0.1" />
```
eriove: Since there was only one test unit MS Test I switched that to NUnit before adding more tests.

Source/HelixToolkit.SharpDX.Shared/Model/Scene/CrossSectionMeshNode.cs
```diff
@@ -238,6 +240,86 @@ protected override RenderCore OnCreateRenderCore()
             {
                 return new CrossSectionMeshRenderCore();
             }
+
+            protected override bool OnHitTest(RenderContext context, Matrix totalModelMatrix, ref Ray rayWS, ref List<HitTestResult> hits)
+            {
+                int hitsBeforeCheck = hits?.Count ?? 0;
+                var meshGeometry3d = Geometry as MeshGeometry3D;
+                if (meshGeometry3d == null)
```
eriove: Is it correct to return false or is it better to cast (and get an exception) or at least to log this?

Source/HelixToolkit.Tests/HelixToolkit.Tests.csproj
```diff
@@ -57,7 +57,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="NUnit">
-      <Version>3.10.1</Version>
+      <Version>3.12.0</Version>
```
eriove: The   `[Apartment(ApartmentState.STA)]` did not exist in the .NET Core version 3.10.1 so I upgraded all to 3.12.0 for consistency.

Source/HelixToolkit.SharpDX.Shared/Model/Geometry/MeshGeometry3D.cs
eriove: I couldn't find a way to keep backwards-compatibility and add more parameters so I added a property instead. That might break if the same `MeshGeometry3D` is used in different locations. Is that a likely scenario?

My guess is that the closest hit test will be taken in most cases anyway, then this wont break the calling code, just make it slower.

Source/HelixToolkit.SharpDX.Shared/Model/Scene/CrossSectionMeshNode.cs
holance: Just return false I think. If it is not MeshGeometry3D, the mesh will not be able to render on screen.

Source/HelixToolkit.SharpDX.Shared/Model/Scene/CrossSectionMeshNode.cs
```diff
@@ -238,6 +240,86 @@ protected override RenderCore OnCreateRenderCore()
...
+                    return false;
+                meshGeometry3d.ReturnMultipleHitsOnHitTest = true;
```
holance: Set it back to false after hit test, since MeshGeometry3D can be shared across multiple models.

eriove: Hmm, in that case there is a risk that one call sets it and then the next one unsets the flag before the first call sets it back. Or are all of these calls guaranteed to be on the same thread?

eriove:
I can see four options how to resolve this at the moment.

1. Add an overload to `MeshGeometry3D.HitTest()` with an argument to return multiple hits. Put this in a new interface and check for that interface in `CrossSectionMeshNode.OnHitTest()`.
2. Add a marker interface and check for that interface in `MeshGeometry3D.HitTest()`.
3. Add an attribute to `originalSource` passed to `MeshGeometry3D.HitTest()`
4. Trust that `CrossSectionMeshNode.OnHitTest()` will always be called from just one thread and set the property back to false when returning.

1 has a better separation of concerns than 2, only `CrossSectionMeshNode` knows why it needs multiple hits returned, `MeshGeometry3D` just returns the additional hits. Might be [slightly slower](https://stackoverflow.com/a/50288277/1144943), but the slower code will only be hit when a `CrossSectionMeshNode` is used.

I think 2 will require the smallest number of changes compared to the current master (`CrossSectionMeshNode.OnHitTest()` can be reverted). It is breaking the [.NET Framework design guidelines](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/interface?redirectedfrom=MSDN) though.

3 is much slower, I only include it since attributes are presented as an alternative to marker interfaces.

4 is easiest (since it is a one line change from the current state of this branch). But given enough objects it doesn't seem unlikely that someone will try to parallelize this code in the future, even if it is safe today.

**What do you prefer?** I prefer 1, but it is a close call with 2.

Source/HelixToolkit.SharpDX.Shared/Model/Scene/CrossSectionMeshNode.cs
holance: Yes, hit test is in UI thread.

holance:
4 should be fine. Unless some one render same geometry multiple times and using cross section on one of them. then also doing parallel hit test. The chance seems to be low. And we can always fix it in separate PR.

eriove:
Sorry for the late reply, I've been ill but are OK again.

Implemented 4, let me know if I should do anything else before merging.

Issue #1419, opened 2020-10-08
Torus class TubeDiameter Property binding fails in WPF UserControl 


ikonuk:
I created a user control containing a TorusVisual3D object inside HelixViewport3D. Everything works. But when I bind to the TubeDimaterProperty in XAML or in code my program crashes. I am not very advanced in WPF. I was not able to pinpoint the exact issue. Since other properties work well and binding in code also fails for the TubeDimaterProperty, I am inclined to think the problem is with the TubeDiameter property. I looked into the source code, as I indicated I am not very advanced in WPF.

`<UserControl x:Class="Torus2020A.TorusUserControl"
             xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
             xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
             xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" 
             xmlns:d="http://schemas.microsoft.com/expression/blend/2008" 
             xmlns:hx="clr-namespace:HelixToolkit.Wpf;assembly=HelixToolkit.Wpf"
             xmlns:local="clr-namespace:Torus2020A"
             mc:Ignorable="d" 
             d:DesignHeight="450" d:DesignWidth="800">
    <Grid>
        <DockPanel>
            <hx:HelixViewport3D x:Name="UCTorusView"
                            ModelUpDirection="0,1,0"
                            ZoomExtentsWhenLoaded="True"  
                            ShowCoordinateSystem="True" 
                            PanGesture="LeftClick">
                <hx:TorusVisual3D x:Name="UCTorus"  
                TorusDiameter="{Binding Path=UCTorusDiameter, RelativeSource={RelativeSource AncestorType=UserControl, AncestorLevel=1}}"
                TubeDiameter="{Binding Path=UCTubeDiameter, RelativeSource={RelativeSource AncestorType=UserControl, AncestorLevel=1}}"
                ThetaDiv="{Binding Path=UCThetaDiv, RelativeSource={RelativeSource AncestorType=UserControl, AncestorLevel=1}}"
                PhiDiv="{Binding Path=UCPhiDiv, RelativeSource={RelativeSource AncestorType=UserControl, AncestorLevel=1}}"
                />
            </hx:HelixViewport3D>
        </DockPanel>
    </Grid>
</UserControl>`

This is the failing XAML (removing TubeDiameter or assigning a value to it works):

This is the binding in Code:

[...]

PR #1424, opened 2020-10-11
Bug Fix for issue #1419 - Binding to TorusVisual3D.TubeDiameter fails if viewport is used in a control

chrkon:
If the HelixToolkit.Viewport is used in a user control and a TorusVisual3D is placed into this Viewport the app stops with an exception when the properties of this user control have a binding to TorusVisual3d.TubeDiameter. All bindings to other properties are working. 

This is a really special use case, but the problem could be reproduced. 

To show the issue the WPF Example "Torus" was changed. An user control was added and then placed into the MainView. Changing the TubeDiameter is done by an animation. Just move the mouse over the right slider. After the animation you can control the diameter directly with the slider.


Source/HelixToolkit.Shared/Geometry/MeshBuilder.cs
```diff
@@ -3021,8 +3021,10 @@ public void AddTorus(double torusDiameter, double tubeDiameter, int thetaDiv = 3
                 this.AddSphere(new Point3D(), tubeDiameter, thetaDiv, phiDiv);
             }
             // If the second Diameter is zero, we can't build out torus
-            else if (tubeDiameter == 0.0)
-                throw new HelixToolkitException("Torus must have a Diameter bigger than 0");
+            else if (tubeDiameter <= 0.0)
+            {
+                // do nothing, because a torus with tube diameter 0 or below is not visible.
```
holance, line 3026: I think keeping the exception is ok. It lets user know the parameter is invalid.

chrkon: I don't think so. Throwing the exception is unexpected for the user because you can set the `ThetaDiv` and the `PhiDiv` parameter to zero. Both values are also invalid, but there is no exception thrown. The Torus just disapears. I think the behaviour would be more consistent if the TubeDiameter parameter will behave in an equal way. 

In my opinion "do not show the torus" should be enough to let the user know that a parameter is wrong. 

I think this exception was the basic reason for @ikonuk to open the issue #1419.

holance: Well, I think all invalid parameters should throw invalid parameter exception. We may want to add the parameter check for all parameters.

chrkon: OK, if all invalid parameters will throw an exception, then the behaviour will be consistent. That is an acceptable solution. I will withdraw my pull request.

holance:
@chrkon You don't need to withdraw this PR. I just want you to leave that exception throwing for now. We can have another PR to fix all parameter check.

Your other changes are okay to be merged.

chrkon:
@holance There are only changes in the examples to reproduce the issue #1419 in this Pull Request. But I can't reopen it. The "Reopen" Button is disabled.  

Should I create a new PR?

PR #1441, opened 2020-11-01
ZoomExtents: confusion between horizontal and vertical fov

mdelanno:
`PerspectiveCamera.FieldOfView` is the vertical field of view (y direction). To calculate the horizontal one, we need to do:

    double hfov = pcam.FieldOfView / viewport.ActualHeight * viewport.ActualWidth

If we have a view in paysage orientation (800 x 600 for example), hfov must be greater than vfov. Let says we have a 45° vertical field of view (I'm using degrees instead of radians to make it more clear):

    hfov = 45 / 600 x 800
    hfov = 0.075 x 800
    hfov = 60

Width and height must be reversed and the variables need to be renamed.


holance (approved):

holance (commented):
Seems like getting a compile error

holance (approved):

Issue #1439, opened 2020-10-27
RenderBitmap with custom size I get black bitmap


miquik:
I’d like to take screenshot of current viewport; I used RenderBitmap extension method and I gave correct result. But resolution is low so I use RenderBitmap method with width/height (is it correct?) but I obtain a completely black bitmap after saving to file

chrkon:
Hi @miquik, I have no trouble to render the bitmap with a higher resolution. 

To reproduce your Issue, I have added some code to the ModelViewer Example in my fork repository. You can find the code in my Fork: [chrkon -helixToolkit Fork, Issue 1439](https://github.com/chrkon/helix-toolkit/tree/issue/%231439_RenderBitmap-Custom-Size)

[Example: ModelViewer/MainViewModel.cs](https://github.com/chrkon/helix-toolkit/blob/6e64a0b5e442703b330908eecb51e841b39620ed/Source/Examples/WPF/ExampleBrowser/Examples/ModelViewer/MainViewModel.cs#L192) 
```
       private void CopyBitmapHiResolution()
        {
            var stopwatch = new Stopwatch();
            stopwatch.Start();

            var bitmap = Viewport3DHelper.RenderBitmap(
                                            this.viewport.Viewport, 
                                            1920.0, 1080.0,
                                            new SolidColorBrush(Colors.White), 1);
            try
            {
                Clipboard.Clear();
                Clipboard.SetImage(bitmap);

                stopwatch.Stop();
                this.ApplicationTitle = string.Format(TitleFormatString, this.CurrentModelPath, 
                    $"(rendered as HD bitmap into clipboard in {stopwatch.ElapsedMilliseconds} ms)");
            }
            catch (Exception)
            {
                this.ApplicationTitle = string.Format(TitleFormatString, this.CurrentModelPath, "(storing in clipboard failed)");
            }
        }
```
I have added the stopwatch to show the delay in rendering with a new resolution.

In the example it takes up to 3,1 seconds to create the HD bitmap. Be sure, that your bitmap is already rendered before you save the file. 

Creating a bitmap with same resolution is fast, because the view is already shown on the screen. Creating a bitmap with a different resolution takes more time, because the whole Model has to be created and rendered a second time.

Kind regards,
Christof

miquik:
Thank you very much for your answer,
I forgot to say that I am using Viewport3DX and I already use "SharpDX" version of your code above. But that gives me completely black bitmap result

chrkon:
[...]

PR #1445, opened 2020-11-09
Issue/#1439 render bitmap custom size sharp dx

chrkon:
I can confirm Bug #1439 . I have added some code to the `FileLoadDemo` Project in the `HelixToolkit.SharpDX` solution to reproduce it. But it is not solved right now.

See comment at the end of this code snippet:
```
        <hx:Viewport3DX
            x:Name="view"
            Grid.Row="1"
            BackgroundColor="#595959"
            Camera="{Binding Camera}"
            CameraMode="Inspect"
            CameraRotationMode="Trackball"
            EffectsManager="{Binding EffectsManager}"
            EnableSwapChainRendering="{Binding SwapChainProperty}"
            FXAALevel="Low">
            <!-- EnableSwapChainRendering="{Binding SwapChainProperty}" is OK (no bug on rendering with different resolution) -->
            <!-- EnableSwapChainRendering="true" fails! (only black screen on rendering with different resolution) -->
```
Of course the `SwapChainProperty` is initialy set to `true`. So the behaviour should be the same.

### DIFF
--- a/Source/Examples/WPF.SharpDX/FileLoadDemo/MainViewModel.cs
+++ b/Source/Examples/WPF.SharpDX/FileLoadDemo/MainViewModel.cs
@@ -15,6 +15,7 @@ namespace FileLoadDemo
     using HelixToolkit.Wpf.SharpDX.Model;
     using HelixToolkit.Wpf.SharpDX.Model.Scene;
     using Microsoft.Win32;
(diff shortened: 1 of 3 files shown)


chrkon:
This PullRequest was replaced by Pull Request #1469

Issue #1461, opened 2020-11-21
RectangleVisual3D do not work (show) when Normal=(1,0,0)


ikonuk:
I came across this strange behavior in a project I am working. I prepared the attached small example to illustrate the problem.

I. Konuk

[rectangletest.zip](https://github.com/helix-toolkit/helix-toolkit/files/5578080/rectangletest.zip)

chrkon:
Hi @ikonuk, thank you for this message.  - @holance : **Unfortunately I have to confirm this bug.** 

But perhaps you can use the following workaround until the bug is fixed:
Instead of `Normal = new Vector3D(1, 0, 0)` you could use `Normal = new Vector3D(1, float.Epsilon, 0)` .
I know that this is not the same, but `float.Epsilon` is defined as `1.401298E-45`. And this is very close to zero. I have checked this with your sample and it works. 

I take this issue and I will take a closer look on it. 

And many thanks to you for answering questions of other users. This is a way to improve such an open source project. 

Kind regards,
Christof

ikonuk:
Christof: 

Thank you very much again. Thank you also for the workaround.

I really like this project. I will try to do my best to help others. I have limited C# and WPF experience. I am a mathematician and engineer and normally code in C++. Please do not hesitate to contact me If there are any tasks or questions along those lines.

Thank you again for the kind words. Happy Holliday season to everybody.

I. Konuk

chrkon:
Dear I. Konuk,
I have found the real reason for this bug. It is really more a mathematical issue than a software bug. 

If you use the normal vector to define the rectangle plane you need a second vector to define a perpendicular direction for one of the edges. Otherwise the orientation for length and width is not clear and we can't construct the rectangle. 

In `RectangleVisual3D` you can define the `LengthDirection` vector. This vector is (1,0,0) by default. And if both vectors (`normal`and `LengthDirection`) are parallel the cross product will come to 0.0 and the calculation of the rectangle mesh fails. [(see the source here](https://github.com/helix-toolkit/helix-toolkit/blob/17fb3bca62253e5c4041f9bca7dcf4ed62bcc47b/Source/HelixToolkit.Wpf.Shared/Visual3Ds/MeshVisuals/RectangleVisual3D.cs#L197))

[...]

PR #1470, opened 2020-11-28
Fix Rectangle3D failure with normal vector (1, 0, 0) #1461

chrkon:
The reason for this bug was the equality of the properties `normal` vector and `LengthDirection`. The default value of `LengthDirection` is (1, 0, 0). So the real reason for failing was, that the normal vector was parallel to the Length direction vector.

For this case I have used the existing extension method [`FindAnyPerpendicular`](https://github.com/helix-toolkit/helix-toolkit/blob/17fb3bca62253e5c4041f9bca7dcf4ed62bcc47b/Source/HelixToolkit.Wpf.Shared/ExtensionMethods/Vector3DExtensions.cs#L28). But I am not sure, if this is a breaking change. Therefore I have added an example project. Please take a look and play around with different normal verctors. 

I have added the new demo into the `WPF.Example` Browser. I have named it `RectangleNormalVector`. This demo is showing the behaviour of the rectangle if we change the normal vector direction. Especially the change from negative Z to positive Z direction   is sometimes unexpected (but not wrong!). 

Kind regards,
Christof


chrkon:
As in issue #1461 mentinoned, this is not a real bug. I will close it, because the actual behaviour is OK.
Perhaps a hint in the description of the normal vector would be helpful.

PR #1472, opened 2020-12-04
Fix 1438: Viewport3DX in Viewbox crash

### LINKED ISSUE #1438: SharpDX - Viewport3DX inside WPF ViewBox crash
Hello,

I have an issue putting the Viewport3DX of HelixToolkit.Wpf.SharpDX 2.13.1 inside a System.Windows.Controls.ViewBox. The ViewPort3DX is flickering and showing the 3D model multiple times. After some seconds the view port crashes followed by this exception:

Screenshot 2020-10-21 122514

This is the content of my main window, the rest of my project is a default .Net Framework 4.7.1 WPF application:

``` XAML
<Grid>
  <Viewbox>
    <hx:Viewport3DX>
      <hx:Viewport3DX.EffectsManager>
        <hx:DefaultEffectsManager />
      </hx:Viewport3DX.EffectsManager>
      <hx:Viewport3DX.Camera>
        <hx:PerspectiveCamera />
      </hx:Viewport3DX.Camera>
      <hx:AmbientLight3D />
    </hx:Viewport3DX>
  </Viewbox>
</Grid>
```

Regards

JeremyAnsel:
Fix #1438 

- Add example


holance:
This change does not work when with DPI scaling.

JeremyAnsel:
Yes, setting the size of the viewport fixes the issue. But I think that when the size is not set, then the minimal size should be used instead of throwing an exception.

holance:
The changes is still not working properly on my high dpi monitor with scaling enabled.

Current implementation tries to avoid any changes in rendering stage due to dpi scaling. Basically we don't really need to "scale" anything in viewport, but to keep everything the same but render at the original resolution to keep the final image crystal clear.

For example, in wpf, if you set DPI scale to 200% in windows with 3840x2160 resolution. Wpf gives the render host with resolution of 1920x1080. And render host renders image at lower resolution, causes low image quality.

The DPI change basically compensates the DPI scaling and still renders at 3840x2160 resolution.

Maybe the DPI scale naming is a little confusing, it intends to compensate the low resolution due to DPI change in Windows.

holance:
Dpi scaling is still not working properly in latest change. 
I put a fix in my PR #1477 to detect if viewport 3dx is in viewbox. 
Will close this PR for now.

JeremyAnsel:
Your solution seems better than mine.
Thanks.

PR #1604, opened 2021-07-06
Update MeshBuilder.cs

xushengbo:
Improve the efficiency of large-scale model generation

### DIFF
--- a/Source/HelixToolkit.Shared/Geometry/MeshBuilder.cs
+++ b/Source/HelixToolkit.Shared/Geometry/MeshBuilder.cs
@@ -3943,31 +3943,20 @@ namespace HelixToolkit.Wpf
             }
 
             int index0 = this.positions.Count;
-            foreach (var p in positionsToAppend)
-            {
-                this.positions.Add(p);
-            }
+            this.positions.AddRange(positionsToAppend);
 
             if (this.normals != null && normalsToAppend != null)
             {
-                foreach (var n in normalsToAppend)
-                {
-                    this.normals.Add(n);
-                }
+                this.normals.AddRange(normalsToAppend);                
             }
 
             if (this.textureCoordinates != null && textureCoordinatesToAppend != null)
             {
-                foreach (var t in textureCoordinatesToAppend)
-                {
-                    this.textureCoordinates.Add(t);
-                }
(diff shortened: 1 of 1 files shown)


Source/HelixToolkit.Shared/Geometry/MeshBuilder.cs
```diff
@@ -3943,31 +3943,20 @@ public void Append(MeshGeometry3D mesh)
             }
 
             int index0 = this.positions.Count;
-            foreach (var p in positionsToAppend)
-            {
-                this.positions.Add(p);
-            }
+            this.positions.AddRange(positionsToAppend);
```
holance, line 3946: AddRange does not available in NetStandard1.3.

Source/HelixToolkit.Shared/Geometry/MeshBuilder.cs
```diff
@@ -3943,31 +3943,20 @@ public void Append(MeshGeometry3D mesh)
...
-                }
+                this.textureCoordinates.AddRange(textureCoordinatesToAppend);             
             }
 
-            foreach (int i in triangleIndicesToAppend)
-            {
-                this.triangleIndices.Add(index0 + i);
-            }
+            this.triangleIndices.AddRange(triangleIndicesToAppend);
+            System.Threading.Tasks.Parallel.For(index0,index0+ triangleIndicesToAppend.Count-1,i=> { this.triangleIndices[i] += index0; } );
```
holance, line 3959: How much performance improvement you have seen by using parallel here?

mbreckon:
@xushengbo I can't speak for @holance but as a PR this is lacking any detail. You've provided a change without any description or documentation of the testing you've done. @holance will need to recreate all of that work and likely run under various different scenarios to ensure there is no serious regression. It is unreasonable to expect that an open source maintainer should pick that up simply because you've created a PR containing a few lines of change. Again I can't speak for @holance but I would think that this might work better as an issue that can be picked up and investigated at a time convenient to the wider project, but if you want it to go into the project on a more reliable schedule then you will likely need to engage significantly more to provide evidence of the testing you've done and support @holance in ensuring there are no significant regressions beyond your specific usage.

holance:
No response from the author.

PR #1644, opened 2021-09-15
Revert "WinUI version (experimental)"

holance:
Reverts helix-toolkit/helix-toolkit#1634


holance:
Will do separate merge to better support WinUI.

PR #1699, opened 2022-01-19
Fixes StLReader for .net5 and above

pedrolamas:
I have been writing an [new thumbnail provider for STL](https://github.com/microsoft/PowerToys/pull/15568) in Windows Explorer, and added it to the Microsoft PowerToys.

It was working fine till there was a shift on the target framework from .netcoreapp31 to .net5.

Since then, the StLReader failed to read any ASCII STL file, and I have now came to the conclusion that the problem is with this check:

https://github.com/helix-toolkit/helix-toolkit/blob/3d353b2ad7d22b421d893e44383f6eac7289c1b6/Source/HelixToolkit.Wpf.Shared/Importers/StLReader.cs#L472-L473

The `line.StartsWith("\0")` bit will work perfectly fine for .netcoreapp31, but in .net5 and .net6 it will always return true!

That is because of a change on how the framework now handles globalization. This is documented here:
 - [.NET globalization and ICU](https://docs.microsoft.com/en-us/dotnet/core/extensions/globalization-icu)
 - https://github.com/dotnet/runtime/issues/50521#issuecomment-811523668

The fix is to just replace `line.StartsWith("\0")` with `line.StartsWith("\0", StringComparison.Ordinal)`, but as that line is also checking for empty string (`Length == 0`), one can optimize it by just checking `line[0] == '\0'`

I've targeted the master branch in the hope that you can accept it and push a patch release on the current version!


pedrolamas:
I've added a new commit to this PR to update the self-signed certificates in the UWP apps, as the existing ones have expired and the build was failing due to that.

Source/HelixToolkit.Wpf.Shared/Importers/StLReader.cs
```diff
@@ -469,8 +469,8 @@ private bool TryReadAscii(Stream stream)
                 }
 
                 line = line.Trim();
-                if (line.Length == 0 || line.StartsWith("\0") || line.StartsWith("#") || line.StartsWith("!")
-                    || line.StartsWith("$"))
+                if (line.Length == 0 || line.StartsWith("\0", StringComparison.Ordinal) || line.StartsWith("#")
+                    || line.StartsWith("!") || line.StartsWith("$"))
                 {
```
paulomorgado, line 473: Consider using:

```csharp
                if (line.Length == 0 || line[0] == '\0' || line[0] == '#' || line[0] == '!' || line[0] == '$')
```

Using `ReadOnlySpan<char>` instead of `string` would also be a nice performance improvement.

pedrolamas: Done!

pedrolamas:
Build is broken possibly due to me doing forced pushes, so will close this PR and create a brand new one.

PR #1798, opened 2022-07-17
Edit Viewcube3D size

MauNguyenVan:


holance:
Seems like you are changing a lot of default values. Since it affects all the existing users who may have different size/color preferences. It's more desirable to make them settable and keep existing default values for backward compatibilities, unless you have a really good reason to change them.

PR #1887, opened 2023-02-02
Add EnableParallelUpdate property

kthyholdt:
I noticed when running parallel tasks in my program that the user interface becomes less responsive and the 3d viewer might stutter a bit. I've located it to be because or the .Wait for tasks in the renderloop. Since all threads are busy doing computational work, there are no available threads to do the tasks requested by Helix.

As you can see from the dotTrace screenshot below, from the 12 seconds that are used to UpdateAndRender. 11 seconds are used to Wait for the tasks to complete. 

image

I've added an EnableParallelUpdate property that can be set to false to do all the updates on the main thread. 

If you want to test this out, you can check out my branch, where I have updated the CustomShaderDemo with an example.

### DIFF
--- a/Source/HelixToolkit.SharpDX.Shared/Render/RenderHost/DefaultRenderHost.cs
+++ b/Source/HelixToolkit.SharpDX.Shared/Render/RenderHost/DefaultRenderHost.cs
@@ -66,6 +66,8 @@ namespace HelixToolkit.UWP
                 FrustumEnabledChanged += (s, e) => { SetupFrustumTestFunctions(); };
             }
 
+            public bool EnableParallelUpdate { set; get; } = true;
+
             /// <summary>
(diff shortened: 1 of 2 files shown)


kthyholdt:
I had to commit the example to a different branch so it did not get committed to this PR, you can find it here:

https://github.com/kthyholdt/helix-toolkit/tree/EnableParallelUpdatesExample

holance:
Could you try this change and test whether it solves your issue? https://github.com/helix-toolkit/helix-toolkit/pull/1888

kthyholdt:
@holance Yeah, your change also seems to fix this issue quite well. I won't be able to test it on my real application before it is a nuget package. But on the test application your solution seemed to give identical improvements as this one.

holance:
Fixed in #1888

PR #1907, opened 2023-02-26
Fix WinUI memory leak due to swapchain native panel not being released properly.

holance:
Fix WinUI memory leak due to swapchain native panel not being released properly.


holance:
Discard and replace with an new one.

Issue #1954, opened 2023-05-03
[Helix WPF 2.23] Wrong selection result in rectangle SelectionHitMode =SelectionHitMode.Inside when apply CuttingPlaneGroup


MauNguyenVan:
https://user-images.githubusercontent.com/48544354/235821643-6bc794b9-a17d-4b6e-954c-41357965b1dc.mp4

## Reproduce

```xaml
<Window x:Class="LearnHelixTookit.Wpf.Views.FixSelectionCuttingPlaneGroup"
        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
        xmlns:local="clr-namespace:LearnHelixTookit.Wpf.Views"
        xmlns:ht="http://helix-toolkit.org/wpf"
        mc:Ignorable="d"
        Title="FixSelectionCuttingPlaneGroup" Height="450" Width="800">
    <Grid>
        <Grid>
            <Grid.RowDefinitions>
                <RowDefinition Height="50"></RowDefinition>
                <RowDefinition></RowDefinition>
            </Grid.RowDefinitions>
            <StackPanel>
                <CheckBox IsChecked="{Binding ElementName=cuttingGroup1, Path=IsEnabled}">Enable Cutting</CheckBox>
            </StackPanel>
            <ht:HelixViewport3D Name="viewport" Grid.Row="1"
                                Camera="{ht:PerspectiveCamera 10,5,5}" 
                                ZoomExtentsWhenLoaded="True" 
                                IsViewCubeEdgeClicksEnabled="True">
                <ht:SunLight/>

                <ht:SortingVisual3D>
                    <ht:RectangleVisual3D Origin="{Binding ElementName=planeManipulator1, Path=Position}" Normal="1,0,0" LengthDirection="0,1,0" Fill="#800000FF"/>
                    <ht:CuttingPlaneGroup x:Name="cuttingGroup1"  >
                        <ht:CuttingPlaneGroup.CuttingPlanes>
                            <ht:Plane3D x:Name="pl1" Normal="1,0,0"  Position="0,1,0"/>
                        </ht:CuttingPlaneGroup.CuttingPlanes>
[...]

PR #1971, opened 2023-05-10
Fixed rectangle selection cause model3D.Bounds is empty, improve performance

MauNguyenVan:
- Remove the magic number in ViewCubeVisual3D [WPF]
- Fixed rectangle selection cause Model3D.Bounds is Empty, improve performance  [WPF]. ref #1954 
- Make BoundingBoxVisual3D into one model [WPF]

### DIFF
--- a/Source/HelixToolkit.Wpf.Shared/Helpers/Viewport3DHelper.cs
+++ b/Source/HelixToolkit.Wpf.Shared/Helpers/Viewport3DHelper.cs
@@ -223,48 +223,65 @@ namespace HelixToolkit.Wpf
                 (model, visual, transform) =>
                 {
                     var geometry = model.Geometry as MeshGeometry3D;
-                    if (geometry == null || geometry.Positions == null || geometry.TriangleIndices == null)
+                    if (geometry == null
+                    || model.Bounds.IsEmpty
+                    || geometry.Bounds.IsEmpty
+                    || geometry.Positions == null
+                    || geometry.Positions.Count == 0
+                    || geometry.TriangleIndices == null
+                    || geometry.TriangleIndices.Count == 0)
                     {
                         return;
                     }
+                    // ref https://learn.microsoft.com/en-us/archive/blogs/wpf3d/transforming-bounds
+                    GeneralTransform3DTo2D transform3DTo2D = visual.TransformToAncestor(viewport);
+                    // If null, the transform isn't possible at all
+                    if (transform3DTo2D is null) return;
+                    Rect visualBound = transform3DTo2D.TransformBounds(VisualTreeHelper.GetDescendantBounds(visual));
 
-                    var status = mode == SelectionHitMode.Inside;
-
-                    // transform the positions of the mesh to screen coordinates
-                    var point2Ds = geometry.Positions.Select(transform.Transform).Select(viewport.Point3DtoPoint2D).ToArray();
-
-                    // evaluate each triangle
-                    for (var i = 0; i < geometry.TriangleIndices.Count / 3; i++)
-                    {
-                        var triangle = new Triangle(
-                            point2Ds[geometry.TriangleIndices[i * 3]],
-                            point2Ds[geometry.TriangleIndices[(i * 3) + 1]],
-                            point2Ds[geometry.TriangleIndices[(i * 3) + 2]]);
-                        switch (mode)
-                        {
-                            case SelectionHitMode.Inside:
-                                status = status && triangle.IsCompletelyInside(rectangle);
-                                break;
-                            case SelectionHitMode.Touch:
-                                status = status
-                                         || triangle.IsCompletelyInside(rectangle)
-                                         || triangle.IntersectsWith(rectangle)
-                                         || triangle.IsRectCompletelyInside(rectangle);
-                                break;
-                        }
-
-                        if (mode == SelectionHitMode.Touch && status)
-                        {
-                            break;
-                        }
-                    }
-
-                    if (status)
+                    if (rectangle.Contains(visualBound)) // object is both inside and touch
                     {
                         results.Add(new RectangleHitResult(model, visual));
                     }
-                });
+                    else if (rectangle.IntersectsWith(visualBound))
+                    {
+                        // transform the positions of the mesh to screen coordinates
+                        var point2Ds = geometry.Positions.Select(transform.Transform).Select(viewport.Point3DtoPoint2D).ToArray();
+                        var meshPoints = new Point[geometry.TriangleIndices.Count];
+                        for (var i = 0; i < geometry.TriangleIndices.Count; i += 3)
+                        {
+                            meshPoints[i] = point2Ds[geometry.TriangleIndices[i]];
+                            meshPoints[i + 1] = point2Ds[geometry.TriangleIndices[i + 1]];
+                            meshPoints[i + 2] = point2Ds[geometry.TriangleIndices[i + 2]];
+                        }
+                        double minX = meshPoints.Min(x => x.X);
+                        double minY = meshPoints.Min(x => x.Y);
+                        double maxX = meshPoints.Max(x => x.X);
+                        double maxY = meshPoints.Max(x => x.Y);
+                        Rect geometryBound = new Rect(new Point(minX, minY), new Point(maxX, maxY));
 
+                        if (rectangle.Contains(geometryBound))// object is both inside and touch
+                        {
+                            results.Add(new RectangleHitResult(model, visual));
+                        }
+                        else if (mode == SelectionHitMode.Touch && rectangle.IntersectsWith(geometryBound))
+                        {
+                            // evaluate each triangle
+                            for (var i = 0; i < meshPoints.Length; i += 3)
+                            {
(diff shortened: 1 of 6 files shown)


Source/HelixToolkit.Wpf.Shared/Helpers/Viewport3DHelper.cs
```diff
@@ -223,48 +223,65 @@ public static IEnumerable<RectangleHitResult> FindHits(this Viewport3D viewport,
...
+                    else if (rectangle.IntersectsWith(visualBound))
                     {
-                        var triangle = new Triangle(
-                            point2Ds[geometry.TriangleIndices[i * 3]],
-                            point2Ds[geometry.TriangleIndices[(i * 3) + 1]],
-                            point2Ds[geometry.TriangleIndices[(i * 3) + 2]]);
-                        switch (mode)
+                        // transform the positions of the mesh to screen coordinates
+                        var point2Ds = geometry.Positions.Select(transform.Transform).Select(viewport.Point3DtoPoint2D).ToArray();
+                        var meshPoints = new Point[geometry.TriangleIndices.Count];
```
holance, line 250: I think you only need to transform all vertices instead of each triangles since multiple triangles may share same vertices.

Source/HelixToolkit.Wpf.Shared/Helpers/Viewport3DHelper.cs
```diff
@@ -223,48 +223,65 @@ public static IEnumerable<RectangleHitResult> FindHits(this Viewport3D viewport,
...
-                        if (mode == SelectionHitMode.Touch && status)
+                        if (rectangle.Contains(geometryBound))// object is both inside and touch
                         {
-                            break;
+                            results.Add(new RectangleHitResult(model, visual));
+                        }
+                        else if (mode == SelectionHitMode.Touch && rectangle.IntersectsWith(geometryBound))
+                        {
+                            // evaluate each triangle
+                            for (var i = 0; i < meshPoints.Length; i += 3)
```
holance, line 270: Here as well, only need to evaluate whether any vertex is inside the rectangle.

holance: Instead of transform all triangles, just transform all vertices, then use triangle indices to get the transformed vertices back during the intersection testing.

holance:
> I also want to edit those classes to struct to improve performance, but I do not know how that can influence other sourcecode
> 
> https://github.com/helix-toolkit/helix-toolkit/blob/ab9fd6a575362a03583374cead782348dc668ba4/Source/HelixToolkit.Wpf.Shared/Geometry/LineSegment.cs#L17
> 
> https://github.com/helix-toolkit/helix-toolkit/blob/ab9fd6a575362a03583374cead782348dc668ba4/Source/HelixToolkit.Wpf.Shared/Geometry/Triangle.cs#L17

Should be ok to change them to struct.

Source/HelixToolkit.Wpf.Shared/Helpers/Viewport3DHelper.cs
holance: `meshPositions` should have count = `geometry.Positions.count`, otherwise you are still duplicating vertex transforms based on triangle indices. For example, if two triangle has indices of `1, 2, 3` and `1, 2, 4`, now you are transforming 1 and 2 vertices twice.

Source/HelixToolkit.Wpf.Shared/Helpers/Viewport3DHelper.cs
holance: That's the problem from cutting plane, it does not remove the vertices after cutting.  And it should only affect the bounding box, not the final result from the selection, since it still does the intersection check with each triangles.

holance: No plans for fixing it. I only maintain sharpdx versions at this moment.

Source/HelixToolkit.Wpf.Shared/Helpers/Viewport3DHelper.cs
```diff
@@ -223,49 +223,143 @@ public static IEnumerable<RectangleHitResult> FindHits(this Viewport3D viewport,
...
+                                    results.Add(new RectangleHitResult(model, visual));
+                                    break;
+                                }
+                            }
+                        }
                     }
                 });
-
             return results;
+            Point[] TransformPoints3dTo2d(Viewport3D viewport3d, Transform3D transform3d, Point3D[] point3ds)
```
holance, line 330: It's better to use `IList<Point3D>` as parameter instead of `Point3D[]` array, otherwise you have to create an array just for passing in the position list.

Source/HelixToolkit.Wpf.Shared/Helpers/Viewport3DHelper.cs
```diff
@@ -223,49 +223,143 @@ public static IEnumerable<RectangleHitResult> FindHits(this Viewport3D viewport,
...
+                Point[] point2ds = new Point[length];
+                for (int i = 0; i < length; i++)
+                {
+                    Point3D transformedPoint3d = transform3d.Transform(point3ds[i]);
+                    Point point2d = viewport3d.Point3DtoPoint2D(transformedPoint3d);
+                    point2ds[i] = point2d;
+                }
+                return point2ds;
+            }
+            Rect GetBound2d(Point[] point2ds)
```
holance, line 342: Same here.

Source/HelixToolkit.Wpf.Shared/Helpers/Viewport3DHelper.cs
```diff
@@ -223,49 +223,143 @@ public static IEnumerable<RectangleHitResult> FindHits(this Viewport3D viewport,
...
+            {
+                double minX = point2ds[0].X;
+                double minY = point2ds[0].Y;
+                double maxX = point2ds[0].X;
+                double maxY = point2ds[0].Y;
+                for (int i = 1; i < point2ds.Length; i++)
+                {
+                    Point pi = point2ds[i];
```
holance, line 350: Use `ref Point pi = ref point2ds[i]` to avoid copy.

Source/HelixToolkit.Wpf.Shared/Helpers/Viewport3DHelper.cs
```diff
@@ -223,49 +223,143 @@ public static IEnumerable<RectangleHitResult> FindHits(this Viewport3D viewport,
...
+                    if (pi.X < minX)
```
holance, line 351: Why not just use `Maths.Min` and `Maths.Max`?

Source/HelixToolkit.Wpf.Shared/Helpers/Viewport3DHelper.cs
```diff
@@ -223,49 +223,143 @@ public static IEnumerable<RectangleHitResult> FindHits(this Viewport3D viewport,
...
 
-                    // evaluate each triangle
-                    for (var i = 0; i < geometry.TriangleIndices.Count / 3; i++)
-                    {
-                        var triangle = new Triangle(
-                            point2Ds[geometry.TriangleIndices[i * 3]],
-                            point2Ds[geometry.TriangleIndices[(i * 3) + 1]],
-                            point2Ds[geometry.TriangleIndices[(i * 3) + 2]]);
-                        switch (mode)
+                        int[] distinctTriangleIndices = geometry.TriangleIndices.Distinct().ToArray();
```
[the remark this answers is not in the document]
holance: Are you having incorrect result without getting actual bounding box after cutting plane?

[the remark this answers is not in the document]
holance: > Yes. I have, i mentioned here [#1971 (comment)](https://github.com/helix-toolkit/helix-toolkit/pull/1971#discussion_r1191606636).

What's the incorrect result you are getting from selection? Even the bounding box is larger with extra vertices, final selection result should be the same.

Source/HelixToolkit.Wpf.Shared/Helpers/Viewport3DHelper.cs
holance: Do you have any performance benchmark for comparison? Also, creating a temporary array and copy everything increases memory usage on large models, I don't think the cost can be justified. What happens if you have a mesh contains 1 million vertices?

Source/HelixToolkit.Wpf.Shared/Helpers/Viewport3DHelper.cs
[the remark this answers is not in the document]
holance: I am confused. Why the selection does not work when it is too long? The selection rectangle is intersecting with their bounding box, right?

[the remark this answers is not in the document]
holance: Yes, but then why the selection didn't work if bounding box is too long? Your implementation will check all triangles, right?

holance (approved):

PR #1989, opened 2023-06-20
Fix use of NotifyCollectionChangedEventArgs in ObservableFastList

SnowyWreath:
The three-argument constructor is meant only for Replace, and it throws an exception if you pass Remove. Use the two-argument constructor for Remove.

System.ArgumentException: Constructor supports only the 'Replace' action. (Parameter 'action')
image


holance:
Good catch.

PR #2017, opened 2023-09-12
Change lock to try lock in DX11ImageSource.

holance:


kthyholdt:
I will test it now. One thing I've noticed is that I often get an Exception in DX11ImageSourceRenderer_OnNewBufferCreated

`                    surfaceD3D.SetRenderTargetDX11(e.Texture.Resource as Texture2D);
`

It manages to recover from that, so the animation continues and the scene is rendered fine. But the ViewCube turns black.
image

holance:
> I will test it now. One thing I've noticed is that I often get an Exception in DX11ImageSourceRenderer_OnNewBufferCreated
> 
> ` surfaceD3D.SetRenderTargetDX11(e.Texture.Resource as Texture2D);`
> 
> image
> 
> It manages to recover from that, so the animation continues and the scene is rendered fine. But the ViewCube turns black. image

Updated the PR, please try to reproduce this issue with the new change.

kthyholdt:
Instead of deadlocks from surfaceD3D.Lock() the program usually crash now. I've pasted the log file here: https://pastebin.com/SQsWhwga

In this log file, there are two successfull "wake ups" from screen off. And on the third one it crashes.
Things i've noticed, is that in the unsuccessfull ones, it never triggers the second event: 

"HelixToolkit.Wpf.SharpDX.Controls.DX11ImageSourceRenderHost: Warning: SurfaceD3D front buffer changed. Value = False, last value True"

It only has the first of these events. 
"HelixToolkit.Wpf.SharpDX.Controls.DX11ImageSourceRenderHost: Warning: SurfaceD3D front buffer changed. Value = True, last value False"

kthyholdt:
To provoke this crash, I have set the "When plugged in, turn off my screen after 1 Minute" in Power & Battery settings.
And then I start the BoneSkinDemo, set it to animate (Solus_Rig|anim_falling_in_air) and wait for the screen to turn off. I've found that I have to wait untill I get a message "No signal" on the screen. They turn black first, but it still takes 30 seconds ish before the "No signal" appears. And thats when SurfaceD3D_IsFrontBufferAvailableChanged is triggered.

image

holance:
The crash exception came from wpf framework core, which we are not able to handle it.

holance:
@kthyholdt Could you help me to test the black viewbox issue? Does this patch fix the issue?

kthyholdt:
@holance I can't really say, since it now crashes with UCEERR_RENDERTHREADFAILURE. I do not get the black viewbox anymore, but I have a feeling that i get the full crash more often. Making me think there is a "timing" issue here, that made the viewbox exception making it more likely to get trough the screen event without crashing.

PR #2022, opened 2023-10-01
[HelixToolkit.Wpf] Update CombinedSelectionCommand

MauNguyenVan:
- Add calculate distance, projection from a point to plane
- Update CombinedSelectionCommand
- Fixed format
- Add Rect3DExtensions


Source/HelixToolkit.Wpf.Shared/Controls/HelixViewport3D.cs
```diff
@@ -123,7 +124,7 @@ public class HelixViewport3D : ItemsControl, IHelixViewport3D
         /// </summary>
         public static readonly DependencyProperty CoordinateSystemHeightProperty =
             DependencyProperty.Register(
-                "CoordinateSystemHeight", typeof(double), typeof(HelixViewport3D), new UIPropertyMetadata(80.0));
+                "CoordinateSystemHeight", typeof(double), typeof(HelixViewport3D), new UIPropertyMetadata(100.0));
```
holance, line 127: Any reason to change the default value?

Source/HelixToolkit.Wpf.Shared/Controls/HelixViewport3D.cs
```diff
@@ -746,7 +747,7 @@ public class HelixViewport3D : ItemsControl, IHelixViewport3D
         /// Identifies the <see cref="ViewCubeHeight"/> dependency property.
         /// </summary>
         public static readonly DependencyProperty ViewCubeHeightProperty = DependencyProperty.Register(
-            "ViewCubeHeight", typeof(double), typeof(HelixViewport3D), new UIPropertyMetadata(80.0));
+            "ViewCubeHeight", typeof(double), typeof(HelixViewport3D), new UIPropertyMetadata(100.0));
```
holance, line 750: Same here, any reason to change default values?

Source/HelixToolkit.Wpf.Shared/Geometry/Plane3D.cs
```diff
@@ -19,12 +19,12 @@ public class Plane3D
         /// <summary>
         /// The normal.
         /// </summary>
-        private Vector3D normal;
+        private Vector3D _normal;
```
holance, line 22: Avoid adding the prefix `_` for private variables.

holance: It's a project wide change, so I would suggest to keep current convention is better for this CR.

PR #2043, opened 2023-11-05
Added ability to create tiled image materials with controlled vertical and horizontal scale factor

Tenchiken80:
I've overloaded MaterialHelper.CreateImageMaterial, adding a method to load image in scaled, tiled form. 
Should be useful with seamless textures.


Tenchiken80:
One can test changes by replacing 
`Material = MaterialHelper.CreateImageMaterial(image);`
with
`Material = MaterialHelper.CreateImageMaterial(image, 0.5, 0.5);`
at line 43 in file Examples\ExampleBrouser\Cloth\Flag.cs then compiling, and launching the ClothDemo example.

PR #2083, opened 2023-12-19
Use system.numerics and HelixToolkit.Maths library 

holance:
Replace Sharpdx.Mathematices with system.numerics and HelixToolkit.Maths library to enable simd operation and avoid collection conversion from mesh builder.


JeremyAnsel:
Hello,
I think the build failed because yours commits comes from a branch with the same name of an existing branch: the name of `helix-toolkit:develop3` and `holance:develop3` is `develop3`.
You can try to create a subbranch of `develop3` in your repository and move the new commits there. So the branches will have different names and will not conflict.

PR #2086, opened 2023-12-22
Support color inner rectangle of RectangleAdorner

JeremyAnsel:
Port #2085 from develop to develop3.


holance:
I have been thinking whether we should continue support Helix toolkit wpf in 3.0. We probably can just leave it in 2.0 branch so we don't have to worry about syncing.

holance:
Extra conversation between system.numerics and media3D vectors introduce extra performance penalties and usage complexity. I don't think developers who use wpf package would even considering moving to use 3.0 wpf package with no benefits.

PR #2105, opened 2024-01-01
Delete ShareFunction at HelixToolkit.Geometry

MauNguyenVan:


JeremyAnsel:
Hello,
I think the build failed because yours commits comes from a branch with the same name of an existing branch: the name of `helix-toolkit:develop3` and `AEC-Technologies:develop3` is `develop3`.
You can try to create a subbranch of `develop3` in your repository and move the new commits there. So the branches will have different names and will not conflict.

holance:
Seems like it's the naming issue. Sometimes it works, sometimes it doesn't.

Source/HelixToolkit.Maths/Plane.cs
```diff
@@ -359,18 +359,18 @@ public static bool IntersectsLine(this Plane p, ref Vector3 p0, ref Vector3 p1,
...
-        /// <param name="p"></param>
-        /// <param name="p0"></param>
-        /// <param name="p1"></param>
-        /// <param name="intersection"></param>
-        /// <returns></returns>
-        public static bool IntersectsLine(this Plane p, Vector3 p0, Vector3 p1, out Vector3 intersection)
-        {
-            return IntersectsLine(p, ref p0, ref p1, out intersection);
-        }
+        ///// <summary>
```
holance, line 362: Any reason for commenting out this? You can delete it if you feel it is unnecessary.

holance (approved):

PR #2112, opened 2024-01-06
Delete BoundingBoxExtensions, rename DataStreamExtension -> DataStreamExtensions

MauNguyenVan:
Delete duplicate code in BoundingBoxExtensions
rename DataStreamExtension -> DataStreamExtensions


Source/HelixToolkit.Maths/BoundingBoxHelper.cs
```diff
@@ -3,33 +3,33 @@
...
-        public static BoundingBox GetBoundingBox(this IList<Vector3> points)
+        public static BoundingBox FromPoints(IList<Vector3>? points)
         {
             if (points == null || points.Count == 0)
             {
                 return new BoundingBox();
             }
-            Vector3 min = new(float.MaxValue);
-            Vector3 max = new(float.MinValue);
+            var min = new Vector3(float.MaxValue);
```
holance, line 19: Why do we want to change this back? I changed it because of the visual studio suggestion from latest C# standard.

Source/HelixToolkit.Maths/BoundingBoxHelper.cs
```diff
@@ -143,5 +143,9 @@ public static BoundingBox Transform(this BoundingBox box, Matrix4x4 transform)
 
             return new BoundingBox(min, max);
         }
+        public static RectangleF Translate(this RectangleF rect, Vector2 translation)
```
holance, line 146: I don't think this function belongs to BoundingBoxHelper. Should be in RectangleF helper or simply put it inside the RectangleF.

holance: Based on input, since it's the extension method of the input type, not the output type.

holance (approved):

holance:
> When do you plan to release the new version?

V2 or V3 release?

PR #2114, opened 2024-01-10
Arrange, restructure code

MauNguyenVan:
Please review my code change. Thanks


Source/HelixToolkit.Maths/Matrix4x4Extensions.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Threading.Tasks;
+
+namespace HelixToolkit.Maths
+{
+    public static class Matrix4x4Extensions
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static Matrix4x4 Inverted(this Matrix4x4 m)
+        {
+            return Matrix4x4.Invert(m, out var result) ? result : Matrix4x4.Identity;
```
holance, line 14: We probably don't want a silent failure if the matrix is not able to be inverted.

holance: Return identity matrix is still a bug. We can log the error if it's not able to be inverted.

holance: You can return either identity or zeros, just log the error.

Source/HelixToolkit.Maths/Plane.cs
```diff
@@ -341,13 +341,14 @@ public static bool IntersectsLine(ref Plane p, ref Vector3 p0, ref Vector3 p1, o
...
             if (d == 0)
             {
                 intersection = Vector3.Zero;
                 return false;
             }
-            float u = (Vector3.Dot(p0, p.Normal) + p.D) / d;
+            float u = (Vector3.Dot(p0, planeNormalize) + p.D) / d;
             intersection = p0 + u * (p1 - p0);
             return true;
         }
```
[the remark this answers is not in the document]
holance: Usually normal means it's already normalized.

Source/HelixToolkit.Maths/Vector3.cs
```diff
@@ -148,6 +144,19 @@ public static bool IsZero(this Vector3 v)
...
 
+        /// <summary>
+        /// Determines whether the specified vector is undefined (NaN,NaN,NaN).
+        /// </summary>
+        /// <param name="v">The vector.</param>
+        /// <returns>
+        /// <c>true</c> if the specified vector is undefined; otherwise, <c>false</c>.
+        /// </returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool IsUndefined(this Vector3 v)
```
ArieRiezebos, line 155: Comment suggests that it returns true if all (X, Y, Z) are NaN, but it returns true if any of them is NaN. I'm also wondering what the difference is with method AnyUndefined() on line 879. What it maybe the intention to use && for this method and || for the other?

holance: Comment is wrong

holance (approved):

holance (approved):

PR #2123, opened 2024-01-18
Update name to more consistency

MauNguyenVan:


Source/HelixToolkit.Maths/Plane.cs
```diff
@@ -60,6 +60,7 @@ public static Plane Create(Vector3 point1, Vector3 point2, Vector3 point3)
             Vector3 p13 = point3 - point1;
             Vector3 normal = Vector3.Normalize(Vector3.Cross(p12, p13));
             return new Plane(normal, -Vector3.Dot(normal, point1));
+            //return Plane.CreateFromVertices(point1, point2, point3);
```
holance, line 63: Is this code redundant?

Source/HelixToolkit.Maths/Plane.cs
```diff
@@ -86,6 +87,7 @@ public static Plane NormalizeUnit(this Plane plane)
             plane.Normal *= length;
             plane.D *= length;
             return plane;
+            //return Plane.Normalize(plane);
```
holance, line 90: Same here.

Source/HelixToolkit.Maths/Plane.cs
holance: Welcome to remove duplicated functions.

PR #2136, opened 2024-02-06
Improve dual depth peeling implementation. Fix #2116.

### LINKED ISSUE #2116: Unexpected rendering with OITRenderMode = DepthPeeling with alpha=0 or alpha=1
Hello,

My app is as described in issue #2115. This time I'm using OITRenderMode = DepthPeeling, but I face other undesired rendering results when alpha approaches 0 or 1:

- When alpha=1 (or, say 0.99), the surface (green in the screenshot below) properly hides others behind it (the blue plane in the screenshot), except when another transparent surface is positioned in front of it (the red plane):

image

This can be bypassed by setting IsTransparent=false when alpha=1, but the rendering below the red plane is quite different, so that it produces a "jump"  when the user smoothly increases the alpha value up to 1. Compare the screenshot above with the one below (IsTransparent=false):
image

- When alpha=0 (or, say 0.01), it leaves a "shadow" below the red plane and above the blue one:
image

Similarly as above, this can be bypassed by setting Visibility=Hidden, but again, the rendering does not look correct and is not continuous when approaching alpha=0. Compare the preceding screenshot with the one below (Visibility=Hidden):
image

Is this behavior graphic card dependent? Is there a proper way to bypass or improve it?

Thanks again!

holance:
Improve dual depth peeling implementation. Fix #2116.

### DIFF
--- a/Source/Examples/WPF.SharpDX/OrderIndependantTransparentRendering/MainViewModel.cs
+++ b/Source/Examples/WPF.SharpDX/OrderIndependantTransparentRendering/MainViewModel.cs
@@ -96,7 +96,7 @@ namespace OrderIndependentTransparentRendering
         {
             set
             {
-                if(SetValue(ref oitRenderType, value))
+                if (SetValue(ref oitRenderType, value))
                 {
                     switch (value)
                     {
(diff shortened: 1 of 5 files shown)


cleosia:
Thanks a lot!

PR #2139, opened 2024-02-07
Fix duplicate model when handle point hittest

MauNguyenVan:


Source/HelixToolkit.Wpf/Helpers/Viewport3DHelper.cs
```diff
@@ -1398,5 +1413,23 @@ public sealed class HitResult
                 return this.RayHit?.VisualHit;
             }
         }
+
+        public override bool Equals(object? obj)
+        {
+            if (obj is HitResult other)
+            {
+                return this.Visual == other.Visual && this.Model == other.Model;
```
holance, line 1421: How about position and normal values? Do we not need to compare those? 
Same model could be intersected by ray multiple times depending on the shape of the mesh. I don't see the issue of returning same model with different hit position and normal values.

holance: Your change seems to be a special use case. If position and normal are different, then the result is valid. You can always filter duplicated ones outside Helix toolkit if you only care about unique visual and model.

holance: Some developers may care about different positions and normals in same visuals.

PR #2177, opened 2024-03-19
Add TypeConverter to collections

### LINKED ISSUE #2174: HelixToolkit v3 load failure with an Exception System.Windows.Markup.XamlParseException in ExampleBrowser
### Reproduce:
- Open ExampleBrowser (Examples\Wpf.SharpDX\ExampleBrowser)
- Select `Issue 1044` `Issue 1048 and 1052` `Issue 1047-1`  `Issue 1047-1`

### Result
- Expect result: Load successful
- Actual result: Load failure with throw an exception `System.Windows.Markup.XamlParseException: `    
And I guess in other places have the same error related to `System.Windows.Markup.XamlParseException`
devenv_8w346iLmEg

JeremyAnsel:
Fix #2174 

Move collection converters from `HelixToolkit.SharpDX` to `HelixToolkit.Geometry`.
Add `TypeConverter` attribute to collections.


JeremyAnsel:
At first, I think they can be moved to `HellixToolkit` because:
- `HellixToolkit` contains a `Collections` folder
- the collections inherit from `FastList` which is implemented in `HellixToolkit`.

But `Color4Collection` depends on `Color4` which is defined in `HellixToolkit.Maths`.
So I think it would be better to move them to `HellixToolkit.Maths`.

JeremyAnsel:
I've moved the collections and type converters from HelixToolkit.Geometry to HelixToolkit.Maths

PR #2181, opened 2024-03-21
Fixed ViewBoxNode wrong color axis face in Helix.Wpf.SharpDX

MauNguyenVan:


Source/HelixToolkit.Wpf/Visual3Ds/Composite/ViewCubeVisual3D.cs
```diff
@@ -692,9 +690,9 @@ void CreateCubeCorners()
 
         if (this.Size == 0) return;
 
-        double halfSize = this.Size / 2;
-        double sideLength = this.Size / 5;
-        var moveDistance = Math.Sqrt(3) * (sideLength / 2 - this.Overhang);
+        float halfSize = (float)this.Size / 2;
+        float sideLength = (float)this.Size / 5;
+        float moveDistance = (float)Math.Sqrt(3) * (sideLength / 2 - this.overhang);
```
ArieRiezebos, line 695: This will result in a slightly decreased accuracy and there is no performance benefit to casting the input to float instead of using doubles. What is your motivation for this?

holance (approved):

PR #2195, opened 2024-04-04
Fix MeshBuilder.GetCircle() OutOfRangeExeption when thetaDiv <=0

MauNguyenVan:


holance:
Throw an argument exception if thetaDiv <=0.

holance:
Throw exception if div <= 1 since it is not valid.

holance (approved):

Issue #2196, opened 2024-04-06
HelixToolkit.MeshBuilder.AddTube() display wrong profile direction


MauNguyenVan:
I create a tube with profile , use method https://github.com/helix-toolkit/helix-toolkit/blob/ec8aa3334974863eb75ca3474e139dd7c7a46d00/Source/HelixToolkit.Geometry/Geometry/MeshBuilder.cs#L3523-L3524

With rectangle section example.
```cs
var builder = new MeshBuilder(true, true);
var profile = new Vector2[] { new Vector2(-3, 2), new Vector2(3, 2), new Vector2(3, -1), new Vector2(-3, -1) };
builder.AddTube(
    new Vector3[] { new Vector3(), new Vector3(0, 0, 1) },
    null, null, profile,
    false, true);

return builder.ToMesh().ToWndMeshGeometry3D(true);
```
see full my example [here](https://github.com/AEC-Technologies/helix-toolkit/blob/develop3/Source/Examples/WPF/ExampleBrowser/Examples/MeshBuilderSample/MainWindow.xaml.cs)

The result displays the incorrect tube section's direction
![image](https://github.com/helix-toolkit/helix-toolkit/assets/48544354/c5bb3fab-3e1c-434d-815b-cf1229a62351)

holance:
This is not a bug, since the tube is created solely depending on the path direction, mesh builder has to randomly figure out an up and right vector from the path segments, then use those two vectors and sections to extrude the tube.

If you really want to control how up/right vectors being generated, you have to modify the code and pass in additional parameters containing up vectors for each segments in paths.

holance:
Try to avoid making 90 degree change on the paths. You need to create several small steps if you need large angle change on the path.

holance:
No doc, just imagine you are bending a tube, there's no way you can bend it 90 degree without any transition region to try to maintain its cross section shape.

holance:
The degree is calculated based on the previous tube direction, which is Z in your case.

PR #2200, opened 2024-04-12
Allow AddTube function to accept optional up vectors for path segments.

holance:
Allow AddTube function to accept optional up vectors for path segments. Ref: #2196


holance:
> with your optional `pathUps` Whether it can create a twist tube like this image? image

I think so. You can try it out.

PR #2227, opened 2024-06-04
Fix Dispose wrong position of RenderTargetBitmap in Wpf.ParticleSystem

MauNguyenVan:


JeremyAnsel:
The error is not related to this PR.
See #2229

Issue #2180, opened 2024-03-20
ViewportCommands.ZoomExtents zooming not correct in Helix.Wpf.SharpDX V3


MauNguyenVan:
Run Wpf.SharpDX\ExampleBrowser at Source\Examples\Wpf.SharpDX\ExampleBrowser\Workitems\Workitem10043\   
Expect: zoom fit all element in the view
Actual: Not display element  

By the way; why do I zoom in, the button `ZoomExtents` is not enable. I don't see any setting for it

![devenv_LWmDlPCbnU](https://github.com/helix-toolkit/helix-toolkit/assets/48544354/b8ce9b7d-7996-4576-bb3b-a11b04e08b62)

PR #2244, opened 2024-06-26
Fix zoom extend. #2180

holance:
Fix zoom extend. #2180


Source/HelixToolkit.SharpDX/Extensions/CameraCoreExtensions.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using HelixToolkit.SharpDX.Cameras;
 using SharpDX;
+using System.Runtime.InteropServices.ComTypes;
```
JeremyAnsel, line 3: You can remove this line

PR #2274, opened 2024-10-15
Remove HitTestResultCallback in Viewport3DHelper.FindHits(), remove nullable types

MauNguyenVan:


Source/HelixToolkit.Wpf/Helpers/Viewport3DHelper.cs
```diff
@@ -205,17 +204,11 @@ public static IEnumerable<RectangleHitResult> FindHits(this Viewport3D viewport,
...
-            {
-                return new List<RectangleHitResult>();
-            }
-
+            IList<PointHitResult> hitResults = FindHits(viewport, rectangle.BottomLeft);
             return hitResults.Select(x => new RectangleHitResult(x.Model, x.Visual));
         }
 
-        var results = new List<RectangleHitResult>();
+        List<RectangleHitResult> results = new List<RectangleHitResult>();
```
holance, line 211: Any reason why we don't want to use `var` here anymore?

holance (approved):

PR #2302, opened 2024-12-06
[V3] Move more elements to sharpdx shared project

JeremyAnsel:


JeremyAnsel:
The build failed in the tests because there are tempory files being written with the same name by multiple tests.
It is fixed in #2299

PR #2305, opened 2024-12-09
[V3] Add sharpdx D2DScreenMenuDemo to WinUI from Wpf

JeremyAnsel:
Hello,
I've moved more code to the sharpdx shared project.
You can now add 2d elements on top of  the 3d viewport in WinUI like in Wpf.

I've added the D2DScreenMenuDemo example from WPF.SharpDX to WinUI.SharpDX.
In this example 2d elements like image, texts, buttons are rendered on top of a 3d scene.


holance:
2D elements are not really useful for winui. Winui uses swapchain for its rendering, there's no performance penalty for using winui xaml on top of the viewport.

JeremyAnsel:
Good to know. Thanks for the explain.

The idea with the 2d elements is to be able to define 2d elements with the same xaml code in Wpf.SharpDX or in WinUI.SharpDX.

With most model elements (3d or 2d) being in the sharpdx shared project there are now almost the same features for the Viewport3DX control when the project is either Wpf.SharpDX either WinUI.SharpDX, with the same xaml.

PR #2346, opened 2025-03-27
feat: add NASA SRTM .hgt format support for terrain rendering

alex-valchuk:
This new feature allows to render terrain models using SRTM HGT files from NASA.
It also scales the mechanism of rendering terrains models that could be implemented by users themselves.
[SRTM](https://github.com/itinero/srtm) library is used to read .hgt files. It is included into the project as a binary because the original one does not contain significant changes allowing to read required data from objects.

As soon as the [refactoring](https://github.com/itinero/srtm/pull/16) is applied by [itinero](https://github.com/itinero) there will be another PR to change the reference from binary to nuget package.
HGT SRTM Rendering


alex-valchuk:
requires some improvements

PR #2350, opened 2025-03-31
Fix tests

JeremyAnsel:


JeremyAnsel:
With this PR the results are fine when using `run-tests.cmd`.
In the current code the tool may throw an exception when the target framework is `8.0-windows` or `6.0-windows`. I have fixed that.

holance (approved):

PR #2388, opened 2025-05-29
Fix text scale in CoordinateSystemNode

### LINKED ISSUE #2387: Wpf.SharpDX.CoordinateSystemModel3D default size of labels is not readable and does not correlate to the size of the axis when using Mode=AbsolutePosition3D
When using Mode=AbsolutePosition3D the text of the axis labels of a CoordianteSystemModel3D cannot be read because the size of the labels is very small compared to the geometry.

Even when setting the property SizeScale=5 the labels are barely readable.

To reproduce the issue you can use the example VolumeRenderingDemo and add a CoordinateSystemModel3D to the ViewPort3DX in MainWindow.xaml:

        <hx:Viewport3DX BackgroundColor="Black"
            ...
            <hx:CoordinateSystemModel3D Mode="AbsolutePosition3D"
                                                                 SizeScale="5"/>
            ...
        </hx:Viewport3DX>

If you run the example the rendering of the CoordinateSystemModel3D looks ok but the labels are very small which can be seen in the screenshot

Image

JeremyAnsel:
Set scale to `0.5f` for relative mode (current behavior) and to `4.0f` for absolute mode (fix).

Fix #2387

### DIFF
--- a/Source/HelixToolkit.SharpDX/Model/Scene/CoordinateSystemNode.cs
+++ b/Source/HelixToolkit.SharpDX/Model/Scene/CoordinateSystemNode.cs
@@ -297,4 +297,19 @@ public class CoordinateSystemNode : ScreenSpacedNode
     {
         return false;
     }
+
+    protected override void OnCoordinateSystemChanged(bool e)
+    {
+        base.OnCoordinateSystemChanged(e);
+
+        if (axisBillboard.Geometry is BillboardText3D labelText)
+        {
+            float scale = Mode == ScreenSpacedMode.RelativeScreenSpaced ? 0.5f : 4.0f;
+
+            for (int i = 0; i < 3; i++)
+            {
+                labelText.TextInfo[i].Scale = scale;
+            }
+        }
+    }
 }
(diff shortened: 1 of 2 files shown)


holance (approved):

Source/HelixToolkit.SharpDX/Model/Scene/CoordinateSystemNode.cs
```diff
@@ -297,4 +297,19 @@ protected override bool CanHitTest(HitTestContext? context)
     {
         return false;
     }
+
+    protected override void OnCoordinateSystemChanged(bool e)
```
ArieRiezebos, line 301: A bit late maybe, but in my opinion this variable should have a clear and unambiguous name. I have no idea what the bool stands for, it is not an EventArgs object. I would suggest a rename.

JeremyAnsel: `e` is `isRightHand`
Renamed in #2389

PR #2454, opened 2025-12-05
Suggestion - added SolutionItems to the solution

ajgorhoe:
This will enable comfortable viewing and editing of development-related files in the IDE used, such as markdown documentation, CI configurations, common project settings in Directory.Build.props, etc.

### DIFF
--- a/Source/HelixToolkit.sln
+++ b/Source/HelixToolkit.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 17
-VisualStudioVersion = 17.7.34003.232
+# Visual Studio Version 18
+VisualStudioVersion = 18.0.11222.15 d18.0
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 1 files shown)


Source/HelixToolkit.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 17
-VisualStudioVersion = 17.7.34003.232
+# Visual Studio Version 18
```
holance: Let's keep the visual studio version as 17. Not everyone has upgraded to the latest.

holance (changes requested):

Source/HelixToolkit.sln
ajgorhoe: Thanks for your comment, I have reverted the version mark by Visual Studio.

I was thinking of doing this before creating the PR, but did not do it as opening in VS 2022 did not cause any issues for me. Of course, this may be different with older VS releases.

holance (approved):

PR #2477, opened 2026-05-02
Fix WinUI Viewport3DX mouse events

joelspadin:
The fact that Mouse3DEventArgs.HitTestResult is nullable implies that 3D mouse events should be raised regardless of whether something in the scene was clicked, and handlers can check HitTestResult to see if the hit test found anything. However, Viewport3DX's OnMouse3DDown and OnMouse3DMove events were only raised if the hit test found something, so it was not possible to handle events for clicking on nothing or moving the mouse without having clicked on something. The OnMouse3DUp event was never raised at all, because it checked for a hit test, but it always reset the hit test result to null first.

This fixes the events so they are raised even if there's not hit, matching the behavior of the Avalonia and Wpf versions.

Also updated the ModelViewerDemo project to test these changes by making it differentiate between drags for camera pan and clicks for toggling the highlight on nodes.


JeremyAnsel (approved):
LGTM

joelspadin:
Thanks!

